PR #17, opened 2023-01-05
add new relay

cercatrova21:


staab:
That short list is only for quick bootstrapping, and will eventually be removed, the rest are pulled from nostr.watch.

PR #26, opened 2023-02-25
use full event when signing with the extension

fiatjaf:
because if the client is building the event with the pubkey A, but the actual key used by the extension is B, using just the `.sig` property of the value returned by `window.nostr.signEvent()` will cause the event to be invalid.


fiatjaf:
another option is to check if the `.pubkey` of the returned event matches the current pubkey and crash if not.

staab:
That all makes sense, one question and I'll merge

staab:
This has been merged, one way or another. Thank you.

PR #27, opened 2023-02-26
Add a "Details" tab to user profile view

igmaat:
This PR proposes the addition of a "Details" tab to the user profile view. 

**Rationale**

A key reason I use Coracle is because it does a great job with the ergonomics around inspecting and managing relays. There is a balance to strike between keeping the UI minimal, and exposing an appropriate amount of Nostr-related technical detail to users. The only place where I kept missing such detail in Coracle is on the user profile view. This new tab attempts to scratch that itch with some basic but useful information alongside user profiles. The NIP05-related information also aims to provide sufficient detail to troubleshoot NIP05 endpoints, and in particular the relay information they might expose.

**Detail**

- show a profile's public key in both hex and npub formats
- display an nprofile, which is constructed on the fly from either the profile's DB relays, or from the NIP05-advertised relays, if the latter is available. **Note**: this feature is disabled with a flag in the component, pending further consideration and/or refinement.
- When a profile's NIP05 identifier is available:
  - show the full NIP05 identifier
  - show the URL of the NIP05 validation endpoint
  - retrieve the NIP05 validation endpoint's response, in order to...
  - display NIP05-advertised relays, if they are configured/available
- copy buttons where useful

Preview (excluding nprofile):

image

### DIFF
--- a/src/views/person/Details.svelte
+++ b/src/views/person/Details.svelte
@@ -1,0 +1,222 @@
+<script lang="ts">
+  import cx from "classnames"
+  import {nip05, nip19} from "nostr-tools"
+  import {last} from "ramda"
+  import {toast} from "src/app/ui"
+  import Anchor from "src/partials/Anchor.svelte"
+  import Content from "src/partials/Content.svelte"
+  import Spinner from "src/partials/Spinner.svelte"
+  import {copyToClipboard} from "src/util/html"
+  import {warn} from "src/util/logger"
+  import {stringToColor} from "src/util/misc"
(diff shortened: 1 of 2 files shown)


staab:
I'm not at all against this, but could you put this in a modal that you can open with one of the buttons in the header instead? Too many tabs becomes overwhelming very quickly. I need to clean up the header buttons at some point, feel free to take a stab at that, but no worries if not.

igmaat:
Fitting it into a modal would make for better aesthetics.

I'll close & resubmit.

PR #31, opened 2023-03-01
optimize deduplication

fiatjaf:
This will deduplicate incoming events before parsing them (yes, it skips `JSON.parse()`, which apparently is a slow beast) and verifying the signature.

I was told the performance effects of this are noticeable at naked eye.

### DIFF
--- a/src/agent/pool.ts
+++ b/src/agent/pool.ts
@@ -263,17 +263,20 @@ const subscribe = async (
       return
     }
 
-    const sub = conn.nostr.sub(filter, {id})
+    const sub = conn.nostr.sub(filter, {
+      id,
+      alreadyHaveEvent: (id) => {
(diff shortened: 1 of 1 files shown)


staab:
Thank you sir

src/agent/pool.ts
```diff
@@ -263,17 +263,20 @@ const subscribe = async (
...
     }
 
-    const sub = conn.nostr.sub(filter, {id})
+    const sub = conn.nostr.sub(filter, {
+      id,
+      alreadyHaveEvent: (id) => {
+        conn.stats.eventsCount += 1
+        let has = false
+        if (seen.has(id)) has = true
+        seen.add(id)
```
fiatjaf, line 272: This code is stupid.

PR #34, opened 2023-03-02
support nip05 addresses on advanced search

fiatjaf:
Also updated nostr-tools, which I think is fine.

### DIFF
--- a/src/views/search/Scan.svelte
+++ b/src/views/search/Scan.svelte
@@ -3,18 +3,40 @@
   import {onDestroy} from 'svelte'
   import {navigate} from 'svelte-routing'
   import {waitFor} from 'hurdak/lib/hurdak'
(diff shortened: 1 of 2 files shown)


src/views/search/Scan.svelte
```diff
@@ -3,18 +3,40 @@
...
-    goToEntity(result.data)
+    handleInput(result.data)
   }
 
-  const goToEntity = entity => {
-    navigate("/" + entity.replace('nostr:', ''))
+  const handleInput = async input => {
+    input = input.replace('nostr:', '')
+
+    if (any(flip(startsWith)(input), ["note1", "npub1", "nevent1", "nprofile1"])) {
```
staab, line 22: Oh no, it's the ramda cage stage

staab (approved):

PR #36, opened 2023-03-03
lightning stuff parsing improvements and dependencies removed

fiatjaf:
remove bolt11, bech32 and js-lnurl dependencies.

    replace with much faster invoice amount parsing.
    fix lnurl encoding/decoding.


src/util/misc.ts
```diff
@@ -340,11 +339,11 @@ export const uploadFile = (url, fileObj) => {
   return fetchJson(url, {method: 'POST', body})
 }
 
-export const hexToBech32 = (prefix, hex) =>
-  bech32.encode(prefix, bech32.toWords(Buffer.from(hex, 'hex')))
+export const lnurlEncode = (prefix, url) =>
+  bech32.encode(prefix, bech32.toWords(utf8.decode(url)))
```
staab, line 343: This is giving me an error:

```
console.log(lnurlEncode('lnurl', 'https://walletofsatoshi.com/.well-known/lnurlp/angelicdigger21'))

Uncaught (in promise) TypeError: Length 112 exceeds limit 90
    at Object.encode (index.js:331:19)
    at lnurlEncode (misc.ts:343:10)
    at verifyZapper (sync.ts:337:17)
```

staab (changes requested):

src/util/misc.ts
fiatjaf: Oops.

PR #38, opened 2023-03-06
make reaction numbers clickable too

fiatjaf:


src/views/notes/Note.svelte
```diff
@@ -392,26 +392,24 @@
         </div>
         <div class="flex justify-between text-light">
           <div class="flex">
-            <div class="w-16">
-              <button class="fa fa-reply cursor-pointer" on:click|stopPropagation={startReply} />
+            <div class="w-16" on:click|stopPropagation={startReply}>
```
Morkowski, line 395: What do you think about changing this element to button? It would improve accessability and semantics a little bit.
`<button class="w-16 cursor-pointer" on:click|stopPropagation={startReply}>
  <span class="fa fa-reply" aria-hidden="true" />`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden#example
https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-noninteractive-element-interactions.md

Morkowski (approved):

src/views/notes/Note.svelte
staab: I'll see what I can do

fiatjaf:
https://github.com/staab/coracle/commit/c60f7bf5fa55689b1ca8ea45ca5e6286522f8c8c

PR #40, opened 2023-03-08
fallback avatar image added

Morkowski:
Empty avatar circle for profiles without profile image looks like error with fetching profile image, so I put placeholder here.

before:
image

after:
image


src/partials/PersonCircle.svelte
```diff
@@ -0,0 +1,8 @@
+<script lang="ts">
```
Morkowski, line 1: Extends ImageCircle, with the same default values. Just adding falback image.

staab:
Thanks, this does need to be done. However, I'd really like to deterministically generate images based on pubkey so we don't have a ton of undifferentiated orange person icons. I believe this is what Damus does with the robot avatars, although I hate those. This can be done similarly to how relay colors are calculated. Maybe masking the coracle logo with a color palette generation algorithm:

Screen Shot 2023-03-08 at 3 54 54 PM

Morkowski:
What do you think about it?

image

obraz

staab:
This has been merged, with a rebase and some modifications. In particular, I re-used the hsl color generation from relays, which makes it much easier to create a pretty color palette. Thanks!

Issue #133, opened 2023-08-14
Add a highlight color when hovering over note action buttons


maglev99:
Current behavior:

When hovering over the note action buttons "reply", "like", and "zap" with a mouse cursor, the buttons do not change visually.

<img width="627" alt="coracle-action-button-hover" src="https://github.com/coracle-social/coracle/assets/112987694/3953e9c5-db41-424f-a087-7fa4a2ea4720">

Suggested behavior:

When hovering over the note action buttons "reply", "like", and "zap" with a mouse cursor, the buttons should be highlighted in the default highlight color of the button when the action has been performed (ie. performed a "like" action after clicking the "like" button).

<img width="654" alt="coracle-action-button-hover-edited" src="https://github.com/coracle-social/coracle/assets/112987694/f02e68f8-1a86-4f12-962c-fbbe5729e11f">

Explanation:

Adding a highlight color when note action buttons are hovered over improves the usability of the UI by clearly showing the action buttons are interactive.

PR #134, opened 2023-08-14
Added highlight color when hovering over note action buttons

maglev99:
Added tailwind hover class to implement enhancement suggestion:

[Add a highlight color when hovering over note action buttons #133](https://github.com/coracle-social/coracle/issues/133)

Hover class added at the end of string to be consistent with how hover classes are added in the rest of the codebase (ie. in NoteContentKind3.svelte file)


staab:
I've looked at this in the past and I don't think it's the right approach. I've released a different approach to dev, although even that might be too much.

PR #160, opened 2023-09-25
Fix issue where page title does not update after closing a modal

katokishin:
Currently, when opening a Profile modal from the Feeds page, the page title is updated to the name of the user.
However, when closing this modal, the page title does not return to "Feeds" (or whatever modal was open prior, e.g. a different profile's username).

By adding a `prevDocTitle` prop when opening a modal, I was able to modify the code so that closing a modal will revert the page title to whatever it was prior to opening it.

### DIFF
--- a/src/app/Modal.svelte
+++ b/src/app/Modal.svelte
@@ -10,6 +10,7 @@
   $: nonVirtual = $stack.filter(m => !m.virtual)
 
   const closeModal = async () => {
+    document.title = modal.getCurrent().prevDocTitle
     modal.pop()
     menuIsOpen.set(false)
   }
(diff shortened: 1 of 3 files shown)


katokishin:
Hmm actually upon additional testing this seems to leave a case where you:
1. Open a Note modal
2. Open a Profile modal
3. Close the Profile modal
4. Close the Note modal
resulting in an undefined title.

katokishin:
It seems I simply missed a few modal.push()es.
Also changing the name of the parameter from "parentTitle" to "prevDocTitle" for clarity.
Ready for review

staab:
Great, one more thing, can you check to make sure `getCurrent()` is not undefined? Also, if you could change the merge target to the dev branch that would be great.

katokishin:
Updated not to change page title if `getCurrent()` or `prevDocTitle` is nullish. (Would it be better to just let it error if the method `getCurrent()` itself is undefined?)

PR #170, opened 2023-10-07
fix nip98 authorization and actually sign the header

fishcakeday:
This will fix the problem with upload to nostr.build not being authenticated as it should.


staab:
I don't know why, but this causes nostr.build to return a 500 error. I merged it, but I've reverted it for now.

PR #182, opened 2023-10-20
NIP96 support

quentintaranpino:
This is a first version to support NIP96 in Coracle. At the moment there is no list to add or remove media providers (and eventually some way to discover new ones). It simply allows to change the media provider in "settings" using the URL that defines NIP96.

image

It has support for delayed_processing and also for direct processing. 

image


src/app/views/UserSettings.svelte
```diff
@@ -48,6 +48,13 @@
           Allows {appName} to authenticate with relays that have access controls automatically.
         </p>
       </FieldInline>
+      <Field label="Media provider URL">
+        <Input bind:value={settings.mediaprovider_url}>
+          <i slot="before" class="fa-solid fa-server" />
+        </Input>
+        <p slot="info">
+          Enter a custom url for {appName}'s media provider server (BETA)</p>
```
staab, line 56: No need to mark as beta, as long as it works

src/engine/media/model.ts
```diff
@@ -0,0 +1,26 @@
...
+            ["ox", string],
+            ["size", string],
+            ["dim",string],
+            ["magnet", string],
+            ["i", string],
+            ["blurhash", string]
+    ],
+    content: string,
+    sig : string,
+}
```
staab, line 26: No need to be so strict with typescript, I prefer to just use `Event` and assert stuff when needed

src/app/views/UserSettings.svelte
staab: Let's call this "Upload Provider URL" rather than media provider, since the user will be downloading images from other servers. The setting/env var can just be called `nip96_url`

src/partials/ImageInput.svelte
```diff
@@ -45,13 +46,12 @@
...
             }
 
-            const result = await uploadToNostrBuild(body)
-
-            // Legacy weirdness
-            for (const {url} of result.data) {
-              value = url
-              onChange?.(url)
+            const result = await uploadToMediaProvider(settings.mediaprovider_url, body)
+            if (result != "" && result != null){
```
staab, line 50: type coercion is fine, `if (result) {`. Nit aside, it looks like this removes the ability to upload multiple images simultaneously.

src/partials/ImageInput.svelte
```diff
@@ -93,7 +93,7 @@
   <Modal mini onEscape={decline}>
     <Content>
       {#if loading}
-        <Spinner delay={0}>Uploading your media...</Spinner>
+        <Spinner delay={0}>Uploading your media using {settings.mediaprovider_url.split("/.well-known")[0] }</Spinner>
```
staab, line 96: use `displayUrl` instead

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
+
+// This function find the url tag in the NIP94 event
+function findUrlTag(tags: [string, string][]): string {
+  for (const tag of tags) {
+    if (tag[0] === "url") {
+      return tag[1]
+    }
+  }
+  return ""
+}
```
staab, line 95: Can be replaced with `Tags.from(event).type("url").values().first()`

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
+
+const getMediaProviderURL = async (url:string): Promise<string> => {
+  
+  let NIP96_json = await Fetch.fetchJson(url, {method: "GET",})
+  console.debug("NIP96_json", NIP96_json)
+  if (NIP96_json.api_url.length > 0){
+    return NIP96_json.api_url
+  }
+  return "";
 }
```
staab, line 105: use `cached` to avoid making this call too often

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
+  const template = buildEvent(27235, {
+    tags: [
+      ["u", url],
+      ["method", method],
+      ["payload", crypto.createHash("sha256").update(JSON.stringify(body)).digest("hex")]
+    ],
   })
+
+  const $signer = signer.get()
+  const event : Event = await ($signer.canSign()
```
staab, line 78: canSign isn't async, no need to await it

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
+  if(response.nip94_event.tags.length > 0){
+    return findUrlTag(response.nip94_event.tags)
+  }else{
+    return ""
+  }
+
+}
+
+// This function prepare NIP98 event
+async function prepareNIP98Event(url: string, method: string, body : FormData) : Promise<Event> {
```
staab, line 67: Dumb nit, but I prefer using `const prepareNip98Event = async (...) => {`

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
+      await new Promise(resolve => setTimeout(resolve, 1000)) // 1 second of delay
+      timeout++
+    }
+  }
+    
+  if(response.nip94_event.tags.length > 0){
+    return findUrlTag(response.nip94_event.tags)
+  }else{
+    return ""
+  }
```
staab, line 62: This can be simplifed to remove the else by using `Tags` and returning null/undefined if no match

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
       Authorization: `Nostr ${btoa(JSON.stringify(event))}`,
-    },
+  }})
+  console.debug("response", response)
+
+  // If the media provider use delayed processing, we need to wait for the processing to be done
+  if(response.processing_url.length > 0){
+
+    //Prepare NIP98 event	for delayed processing. TODO. If the delayed processing is too long, the event "created_at" maybe expire. It depends on the media provider.
+    const event_processing : Event = await prepareNIP98Event(response.processing_url, "GET", body)
```
staab, line 28: variables use cameCase, attributes use snake_case

src/engine/media/commands.ts
```diff
@@ -1,27 +1,105 @@
...
+        }})
+
+      if (processing_response.status == "success"){
+        console.debug("Processing done");
+        return findUrlTag(processing_response.nip94_event.tags)
+      }
+      console.debug("Status", processing_response.status, "(" + processing_response.percentage + "%)")
+      await new Promise(resolve => setTimeout(resolve, 1000)) // 1 second of delay
+      timeout++
+    }
```
staab, line 55: Hurdak has a nice utility function called `poll` for this kind of thing

staab (changes requested):
Thanks for the PR! I have a few things to change, in comments and below:

- Go ahead and run `yarn format && yarn check` and fix any errors
- Use yarn instead of npm, remove package-lock.json

src/engine/media/commands.ts
quentintaranpino: I thought about it but not really, maybe someone will set up the NIP94 event with the tags in another order, if we leave it like that we will have less chance of error.

src/partials/ImageInput.svelte
quentintaranpino: I'm going to see if I can do something with this

src/partials/ImageInput.svelte
quentintaranpino: Using displayUrl it looks like this:

image

The other way it looks like this:

image

Do you want me to add something like "displayHostname" in misc?

staab: Ah, you're right, I misread the regex. Yeah, that would be great.

src/engine/media/commands.ts
staab: I'm not following? The two should be equivalent.

src/partials/ImageInput.svelte
quentintaranpino: You have displayDomain in "misc", I didn't see it, I will use this one :)

src/engine/media/commands.ts
quentintaranpino: I didn't read you right, I thought you did something like tags[0] 😅😅😅😅😂.

I change it

src/engine/media/commands.ts
quentintaranpino: Done

src/engine/media/commands.ts
quentintaranpino: done

src/engine/media/commands.ts
quentintaranpino: done

src/app/views/UserSettings.svelte
quentintaranpino: done

src/engine/media/model.ts
quentintaranpino: removed

src/partials/ImageInput.svelte
quentintaranpino: done

src/engine/media/commands.ts
quentintaranpino: Without await doesn't work.


This was already in the repository before the commit
image

src/engine/media/commands.ts
quentintaranpino: I have changed it, but there should be an output with "" if the response does not contain a NIP94 event (server malformed for example). 

image

src/engine/media/commands.ts
quentintaranpino: done

src/partials/ImageInput.svelte
quentintaranpino: done

src/engine/media/commands.ts
quentintaranpino: @staab Is it all right if I take it as resolved?

src/engine/media/commands.ts
quentintaranpino: @staab it is ok for you? 😀

src/engine/media/commands.ts
quentintaranpino: done

src/engine/media/commands.ts
quentintaranpino: I can't quite find how you would like it to work, I've left a comment above a test that I don't quite like. 

Is there any documentation on this method?

staab:
Reworked a few things, take another look through and let me know if it all looks good still

quentintaranpino:
> Reworked a few things, take another look through and let me know if it all looks good still

I would have left some debug message (for the percentage of delayed_processing) but it's fine for me.

You are a great developer @staab

quentintaranpino:
I have been talking to fischcake, I asked him how NIP96 is going and he told me that in a few days he will have it ready.

It seems to me that the most honest thing to do is to leave nostr.build by default in the .env file and in the description put a link to other compatible options, like nostrcheck and the rest of its instances.

What do you think?

staab:
> It seems to me that the most honest thing to do

Yes, I think so, but adding a MultiSelect for that field with nostr.build and nostrcheck.me (and any other compliant implementations) available in the options list would also be good.

quentintaranpino:
Actually NIP96 allows you to publish on nostr your list servers. It makes sense to do so.

While waiting for nostr.build I'll give it some thought.

quentintaranpino:
@staab what do you thing about this:

https://nostrcheck.me/media/quentin/1fed5c324fbf8320c9c33f66204bf3910fed46d43afb83aef2f66ae20b6daa4c.mp4

quentintaranpino:
If we do this, the clients should filter out notes with links (to different servers) but with the same filename (the original hash). 

This would serve to give more life to the linked notes, since if one server stops working the rest will serve the data.

staab:
I don't think that's an appropriate burden to put on clients, and would significantly degrade UX. Is there a way, maybe using NIP 94 to create a single address (event id) that points to multiple alternative urls?

quentintaranpino:
Tomorrow I will do it, I was wondering if in NIP94 I could put several x tags

It would be the perfect solution.

I will let you know

quentintaranpino:
Hello @staab, I have uploaded a version that fixes some problems and implements the first version with NIP94. At the moment I just create the event without publishing it and show it in the console, then I extract the first "url" tag of the event to use it in the note (provisional solution).

https://nostrcheck.me/media/quentin/768dee64b29acd3296b9753699cf0281bc74efae64fdf1ddfb6d5da9a03c98a7.mp4

We should think about how to continue, do we publish the NIP94 event? And then we prepare a method to process this event (its URL tags) sequentially until we find one that works?

How do you think it should continue?

quentintaranpino:
I have continued to work, I have been confirmed in the Telegram group of nostr that NIP-94 supports multiple URL's.

Now when uploading an image, we prepare a NIP-94 event for each one and it is added to the note as a quoted nevent.

They can be added and removed with the same functionality as before, but instead of URL's they are nevents. They can be uploaded to several servers at the same time, a single NIP-94 event is published with all the URL's. 

It only remains to solve the fact that Coracle (and probably other clients) show the NIP-94 events in the feed, and therefore when publishing an image (or several), the note is preceded by several NIP-94 notes with each of the images. 

I think the latter is a small but easily solvable drawback.

quentintaranpino:
I have added a selector in the options to use or not use NIP94 events in the notes. This means that if NIP94 is not used, only the first URL returned by the selected set of servers will be used.

image

This provides resiliency but does not take full advantage of NIP96. I have seen that NIP94 is not fully supported (equally) by all other clients and there should be more consensus before publishing files like this.

As it is it is a good middle ground, it allows the user to choose media provider, allows you to have more than one configured and even allows you (if you want) to publish everything under the NIP94 spec. 

When you have time to check it out.

staab:
Great, I think this approach makes sense! I've opened a PR to your branch which rebases and squashes everything onto dev: https://github.com/quentintaranpino/coracle/pull/1/files. I probably broke some stuff, just so you know.

I don't think it makes sense to ask the user to make the nip94 decision for us, since that's basically what they're doing when they upload to multiple servers. I do think publishing nip94's separately could be ok, since that means the image will get picked up both by image-first and general clients (I'll remove 1093 as a default event kind if we do this).

Is it necessary to merge and re-publish the nip94's though? It would be nice to simply add the urls to the kind 1 event directly rather than have to quote the images in. Maybe this is an omission in nip96/94 — the ability to source a single image from multiple urls. You'd know better than me, thoughts?

staab:
@quentintaranpino I haven't forgotten about this, hoping to go over it and get it merged this week or next. We can also add void.cat to our list: https://github.com/nostr-protocol/nips/pull/547#issuecomment-1819274964

quentintaranpino:
> ¡Genial, creo que este enfoque tiene sentido! He abierto un PR a su rama que reorganiza y aplasta todo en dev: https://github.com/quentintaranpino/coracle/pull/1/files. Probablemente rompí algunas cosas, para que lo sepas.
> 
> No creo que tenga sentido pedirle al usuario que tome la decisión de nip94 por nosotros, ya que eso es básicamente lo que están haciendo cuando suben a varios servidores. Creo que publicar nip94 por separado podría estar bien, ya que eso significa que la imagen será recogida tanto por los clientes que dan prioridad a la imagen como por los clientes generales (eliminaré 1093 como tipo de evento predeterminado si hacemos esto).
> 
> Sin embargo, ¿es necesario fusionar y volver a publicar los nip94? Sería bueno simplemente agregar las urls al evento tipo 1 directamente en lugar de tener que citar las imágenes. Tal vez esto sea una omisión en nip96/94: la capacidad de obtener una sola imagen de varias URL. Tú lo sabrías mejor que yo, ¿pensamientos?



> Great, I think this approach makes sense! I've opened a PR to your branch which rebases and squashes everything onto dev: https://github.com/quentintaranpino/coracle/pull/1/files. I probably broke some stuff, just so you know.
> 
> I don't think it makes sense to ask the user to make the nip94 decision for us, since that's basically what they're doing when they upload to multiple servers. I do think publishing nip94's separately could be ok, since that means the image will get picked up both by image-first and general clients (I'll remove 1093 as a default event kind if we do this).
> 
> Is it necessary to merge and re-publish the nip94's though? It would be nice to simply add the urls to the kind 1 event directly rather than have to quote the images in. Maybe this is an omission in nip96/94 — the ability to source a single image from multiple urls. You'd know better than me, thoughts?

If I'm not mistaken NIP94 was created just for this, you quote the NIP94 event in a note and the nostr client must iterate all the URL tags until it finds one that works. 

Anyway, we could solve it directly in NIP94 if we get a way to exchange files between file servers, it's something we've talked about a few times with Fishcake.

Instead of adding several URLs to the NIP94 event, you simply look for the ox tag to the NIP96 servers configured in the client. The first one to respond with the URL is then used in the note. 

I think this is a step further to be agreed in an addition to NIP94 + NIP96, for now you could disable the NIP94 option (just disable) and therefore Coracle will only add the URL of the first NIP96 server that works (eventually it will always be the first one you have in the list). 

I'm fine with adding void.cat, of course. 

Are you ok with this solution?

staab:
> If I'm not mistaken NIP94 was created just for this, you quote the NIP94 event in a note and the nostr client must iterate all the URL tags until it finds one that works.

Ah, that makes sense, I think that would work well.

PR #238, opened 2024-01-19
Revise sidebar styles

dtonon:
See commits for details

image


staab:
@dtonon added a commit that changes how toggleTheme works (subscribe gets called when you update theme), and removes semicolons..

dtonon:
@staab 

>  changes how toggleTheme works (subscribe gets called when you update theme)

That was my first solution, but this way it does not remember the dark theme, every reload it starts with the light one.

> and removes semicolons 

You will have a lot of pain reviewing my js code haha, thanks for the patience :)

> It also removes the one-off styles (e.g. text-[2.6em]). That kind of thing makes it hard for me to be consistent, if there's a specific style we should usually name it in tailwind.config.js rather than use it inline.

Yes I understand. But sometimes one-off adjustments are necessary, and moving them to a dedicated class it seems to me to make all more complex, less Tailwind style, and hidden; I would create a custom class only when it used at least 2 times; keeping these free styles in the code make them obvious to future refactors/merge. Make sense?

staab:
Finalllllyyy getting to this — yep, sounds good, I've added it back in.

PR #261, opened 2024-02-06
fix response from server

quentintaranpino:
In some update Coracle stopped using only the url:

Tags.from(nip94_event).type("url").values().first()

To use all tags, that broke compatibility with all NIP96 servers except those that don't use delayed processing.


staab:
Thanks!

PR #283, opened 2024-02-19
Reduce bottom bar + full black bg in dark mode

dtonon:
The horizontal white line at the bottom is placed by the OS.

image


staab:
Cherry-picked this over.

PR #288, opened 2024-02-26
Retouch mobile navbar

dtonon:
Add back rounded top corners
Use a black border for the profile
Remove border from the search icon

image


staab:
Rebased

PR #293, opened 2024-03-01
Format time and date based on browser regional settings

Morkowski:
Format time and date based on browser regional settings.

For example:
Before - dates were formatted in US style for everyone:
obraz

After - locale read from browser, for my country and language formatting will looks like:
obraz

Solution based on https://stackoverflow.com/a/57529410


staab:
Thank you!

Issue #264, opened 2024-02-08
Bring back delete


staab:
Just make sure to communicate clearly what it does. "Delete" should open a dialog that asks the user to confirm that they want to "request deletion" with a disclaimer that it might not be honored by all relays.

https://coracle.social/notes/nevent1qy2hwumn8ghj7unnwvhxummn9eek7cmfv9kz7qgwwaehxw309ahx7uewd3hkctcpzamhxue69uhhyetvv9ujumn0wvh8xmmrd9skctcpzpmhxue69uhk2tnwdaejumr0dshszxrhwden5te0vdex2ct5wghxummnw3ezuamfdejj7qgewaehxw309aex2mrp0yh8xmn0wf6zuum0vd5kzmp0qqs0r45n9apjnu6m7x2xnkx0wrn8emtu04hhtuvsc3rnw97z7wf2jwq4ha4zy

Tasks:

- [ ] Add delete to overflow menu for an event (only if it was published by the user)
- [ ] Add note/id/create route and dialog
- [ ] Explain that it's only a deletion request and might not be honored by all relays
- [ ] Persist the delete to the user's write relays and to any relays the event was seen on

PR #377, opened 2024-06-30
bring back delete

SamSamskies:
resolves https://github.com/coracle-social/coracle/issues/264

### DIFF
--- a/src/app/views/NoteDelete.svelte
+++ b/src/app/views/NoteDelete.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import Anchor from "src/partials/Anchor.svelte"
-  import Content from "src/partials/Content.svelte"
+  import FlexColumn from "src/partials/FlexColumn.svelte"
   import {router} from "src/app/util/router"
   import {deleteEventById} from "src/engine"
 
@@ -15,7 +15,7 @@
   }
 </script>
 
-<Content size="lg">
+<FlexColumn>
   <p class="flex items-center gap-4 text-xl">
     <i class="fa fa-triangle-exclamation" /> Are you sure you want to delete this event?
   </p>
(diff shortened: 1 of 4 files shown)


src/engine/commands.ts
```diff
@@ -647,8 +647,12 @@ export const publishCommunitiesList = addresses =>
...
 // Deletes
 
-export const publishDeletion = ({kind, address = null, id = null}) => {
-  const tags = [["k", String(kind)]]
+export const publishDeletion = ({kind = null, address = null, id = null}) => {
+  const tags = []
+
+  if (kind) {
+    tags.push(["k", String(kind)])
+  }
```
SamSamskies, line 655: Do you prefer to always include the kind? I didn't see this in the spec so was wondering what's the use case for including it.

src/app/views/NoteDelete.svelte
```diff
@@ -0,0 +1,27 @@
+<script lang="ts">
+  import Anchor from "src/partials/Anchor.svelte"
+  import Content from "src/partials/Content.svelte"
+  import {router} from "src/app/util/router"
+  import {deleteEventById} from "src/engine"
+
+  const onCancel = () => router.pop()
+
+  const onConfirm = () => {
+    deleteEventById($$props.eid)
```
staab, line 10: I prefer `export let eid` when possible

src/app/views/NoteDelete.svelte
```diff
@@ -0,0 +1,27 @@
...
+    router.pop()
+  }
+</script>
+
+<Content size="lg">
```
staab, line 15: Content is deprecated, use `FlexColumn` instead

src/engine/commands.ts
staab: Yeah, there's an open PR: https://github.com/nostr-protocol/nips/pull/1141/files it's just so deletes can be requested more efficiently.

staab (changes requested):
Looks good aside from some minor things, I'm on vacation so it'll probably sit in limbo for a couple weeks before I merge and get back to you with the next thing.

PR #407, opened 2024-08-16
Updates to the code to handle the new event type 1311 Zapstream and nests) 

### LINKED ISSUE #380: Show comments in Coracle
Zapstream and nests use kind 1311. Add them to noteKinds, and in feed.ts only show them if there's a parent (in `buildFeedChunk`).

LeParrain0:
Resolves #380.
I’ve made some updates to the code to handle the new event type 1311 (Zapstream and nests) as requested. Here’s a quick rundown of what’s been changed:

    Note Types Update:
        index.ts: Added type 1311 to the notification kinds.
        state.ts: Included type 1311 in the notifications list.
        index.ts (loadGroupMessages): Updated message filters to include type 1311 for group and community messages.
        feed.ts: Added type 1311 to the default filters.

    buildFeedChunk Function:
        Modified it to ensure that events of type 1311 are only shown if they have a parent. Events without a parent will now be excluded from the feed.


staab (changes requested):
Thanks! A couple changes:

- Instead of adding 1311 in so many places, just add it to the definition of the `noteKinds` array.
- Import `LIVE_CHAT_MESSAGE` from `@welshman/util` instead of using the numeric literal

PR #409, opened 2024-08-16
Use extension for bunker communication

oren-z0:
Instead of signing/encrypting/decrypting the events with a temporary random keypair, we can use the nip07 extension (if installed).

This gives the advantage of not having to re-approve a new app if you logged-out and logged-in (suppose you don't want to keep the session open in your browser). You can also compare the npub of your extension with the npub that requests the permissions (in the nsec.app UI or the nsecbunker.com UI).

Since NIP07 extensions don't have an api to get the private key (and rightfully so!), I've changed the session objects and the NostrConnectBroker to use a connectSession object instead of a connectKey - which has everything you need to sign/encrypt/decrypt the communication thanks to `getSigner` and `getNip04` that were already implemented.

Tested with [nsec.app](https://nsec.app) and [nos2x](https://github.com/fiatjaf/nos2x) on Chrome.

Implementation remark: `NostrConnectBroker`'s implementation calls `getSigner` and `getNip04` once in the constructor and not on-demand. This is because we have two signers/nip04s running here, one for extension and one for the remote-profile. The `getSigner` and `getNip04` memoizations keep only one value in the cache, so calling `getSigner`/`getNip04` on-demand will hit a lot of missing cache values.

### DIFF
--- a/src/app/MenuMobile.svelte
+++ b/src/app/MenuMobile.svelte
@@ -7,7 +7,7 @@
   import PersonCircle from "src/app/shared/PersonCircle.svelte"
   import PersonHandle from "src/app/shared/PersonHandle.svelte"
   import MenuMobileItem from "src/app/MenuMobileItem.svelte"
-  import {globalFeed, slowConnections, menuIsOpen} from "src/app/state"
+  import {slowConnections, menuIsOpen} from "src/app/state"
   import {router} from "src/app/util/router"
   import {makeFeed, normalizeFeedDefinition} from "src/domain"
   import {
@@ -46,8 +46,9 @@
   }
 
   const openFeeds = () => {
-    globalFeed.set(makeFeed({definition: normalizeFeedDefinition(makeScopeFeed(Scope.Follows))}))
-    router.at("notes").push()
+    const feed = makeFeed({definition: normalizeFeedDefinition(makeScopeFeed(Scope.Follows))})
+
+    router.at("notes").cx({feed}).push()
     closeMenu()
   }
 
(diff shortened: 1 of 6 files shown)


staab:
I get the motivation, but I'm not sure 1. when this would happen or 2. if this is a good idea. If I have an extension, wouldn't I just use it to log in directly? Using it as a bunker session that could be shared across apps seems like an extremely unusual approach, and not likely to be widely supported anyhow. Also, if the user is in fact logged in with their primary key to the extension, they would be using their normal key to request signatures for their normal key, which makes no sense, but is probably what would happen most often.

oren-z0:
@staab Think about a profile that can be controlled by multiple users, i.e. a company account. Each employee has his own nsec in a browser extension (preferably encrypted with a local password), but the employee never sees the company's nsec. If one of the employee's computers is lost, or the employee leaves the company, his nsec can be revoked - but there is no need to re-approve him every time he logs out and logs back into the app.

Also, when using a bunker like nsec.app, it is confusing to see a request for permissions from an unknown npub (which could be some man-in-the-middle). I want to approve permissions to a specific npub, and maybe give a list of npubs in advance.

staab:
Yep, I moved all the signer stuff to Welshman, which actually would make more sense to support this kind of configurability. I agree passing a signer instead of a secret to the nip 46 signer would be an improvement (other than the singleton behavior, but I'm sure we could find another way to handle that). I'm still not sure it would make sense to do this within Coracle without having a specific workflow for it. Is this something you're specifically trying to do right now, or is it just an idea?

Issue #397, opened 2024-07-29
Fix wot hover on mobile


staab:
nevent1qyghwumn8ghj7mn0wd68ytnhd9hx2tcpzdmhxue69uhhqatjwpkx2urpvuhx2ue0qywhwumn8ghj7mn0wd68ytndw46xjmnewaskcmr9wshxxmmd9uq32amnwvaz7tmjv4kxz7fwv3sk6atn9e5k7tcppemhxue69uhkummn9ekx7mp0qythwumn8ghj7un9d3shjtnwdaehgu3wvfskuep0qythwumn8ghj7un9d3shjtnswf5k6ctv9ehx2ap0qqsqvvh479k3kuac0sqfy5598j0p978ey7nl93afcjetyvkn384w4gc8spyyt

PR #414, opened 2024-08-31
Fix WoT hover on mobile

zmjohnso:
This PR addresses issue #397.

I changed the `triggerType` in `Popover.svelte` to default to "click" when on mobile, while ignoring clicks to the popover portion of the `PersonName.svelte` component when on mobile to be able to see the WoT score without opening the profile note section. After adding the top-level onClick to the PersonName div, I also changed it to a button to make a11y happy. I think this is fine to do?

I tested locally on mobile and desktop and it seems to work as I'd expect, but I am still new to the project.


staab:
Hey Zachary, just wanted to let you know I haven't forgotten about this, I'm just knee-deep in a refactor at the moment. I'll probably have that wrapped up shortly. In the meantime, no need to fix my merge conflicts, I can handle those.

Also, do you have a npub so I can zap you for your work?

staab:
Rebased and merged, thanks!

Issue #254, opened 2024-01-29
Show scroll bar in modals


staab:
Depending on browser, it can get obscured by the background color we add.

staab:
This ended up leaving artifacts on the right side of the modal on mobile, where you could see the page behind the modal.

PR #424, opened 2024-09-22
Issue 254: Proposed fix

Psycarlo:
Proposed fix for: https://github.com/coracle-social/coracle/issues/254

Is this the desired effect?

One problem is that not every scrollbar has 17px width (https://codepen.io/sambible/post/browser-scrollbar-widths)


staab:
Looks good on MacOS, I'll test on other platforms as I have opportunity, thanks!

PR #435, opened 2024-10-10
Add calendar view for single entity based on new path with npub

### LINKED ISSUE #434: Feature Request: Load calendar events for npub
I love many of the Coracle's features, such as Calendar. I've noticed you can save url for specific event like `coracle.social/events/naddr:...` , this gave me an idea.

If some organization is posting events (real events like meetup and stuff), then it would be extremely helpful to be able to navigate and list directly that org's events in the calendar.

like `coracle.social/events/npub...`

currently it doesn't support npub as url param

( working version in #435 )

tohrxyz:
# Motivation and problem
Hello, as I've wrote in https://github.com/coracle-social/coracle/issues/434 , I really love the calendar mini-app in Coracle client and find it very useful for our organization.

But we also have a specific usecase, where we publish events (meetups and so on..), but need a single page for displaying ONLY OURS calendar events.

I didn't find such option in current coracle client, it can only parse single calendar event from url `coracle.social/events/naddr` , but I propose new additional possibility, where the client could also show calendar events based on specific `npub`.

Such that when I wanted to look at events that `Organization XY` is doing next month, I could go on `URL` and it would show me them - filtered.

# Proposed solution (that is done in commits)
First of all, let me just say that I don't have experience with Svelte for development of web apps, but I am experienced in React side of things. So I studied repo for a bit and found this solution.

1. I've added and registered new route `/events/people/:entity`
2. added check in `Calendar` componet, that if the route is `/events/people/` it should only show notes from that npub

image

# Testing
I've tested the client and it doesn't break anything, normal calendar from UI works like before, but there's a new possibility to go from url path directly to single-entity-mode calendar.

I appreciate all of your work on Coracle and would really like to help with this. If you see any possibilities of improvement to the code, I'm open to it. Thanks

Fixes #434

### DIFF
--- a/src/app/App.svelte
+++ b/src/app/App.svelte
@@ -132,6 +132,12 @@
       address: asNaddr("address"),
     },
   })
+  router.register("/events/people/:entity", Calendar, {
+    required: ["pubkey"],
+    serializers: {
+      entity: asPerson,
+    },
+  })
   router.register("/events/:address/edit", EventEdit, {
     serializers: {
       address: asNaddr("address"),
--- a/src/app/views/Calendar.svelte
+++ b/src/app/views/Calendar.svelte
@@ -4,9 +4,12 @@
   import Calendar from "src/app/shared/Calendar.svelte"
   import {env, loadCircleMessages, userFollows} from "src/engine"
 
+  const isCalendarForSingleEntity = window.location.pathname.startsWith("/events/people/")
   const filter = env.FORCE_GROUP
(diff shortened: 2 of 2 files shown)


staab:
This is probably out of scope, but there are a few existing ways that may work to solve this:

- Post your calendar event to a nip 72 community, and view the community calendar instead of a calendar for a single pubkey
- Create a custom feed that fetches calendar events. This will show the events as cards, rather than on the calendar view, but it might suit the purpose.

Otherwise, something that is more focused on calendar events, like flockstr, might be a better solution.

Issue #388, opened 2024-07-19
Get rid of displayList


staab:
Use new `Intl.ListFormat(getLocale(), {style: "long", type: "conjunction"})` instead. Stick it in a utility function.

zmjohnso:
Been thinking about this one. Using `Intl.ListFormat()` feels like a good idea and would work well but alone it does not cover the same functionality as your `displayLists` library function. Notably this part of the logic: 
```typescript
else if (xs.length > n + 2) {
    return `${xs.slice(0, n).join(", ")}, ${conj} ${xs.length - n} others`
  }
  ```
  Utilizing `Intl.ListFormat()` the updated logic could look something like this:
  ```typescript
export default <T>(xs: T[], conj = "and", n = 6, locale = "en-US") => {
  if (xs.length > n + 2) {
    const formattedList = new Intl.ListFormat(locale, { style: "long", type: "conjunction" }).format(xs.slice(0, n));
    return `${formattedList}, ${conj} ${xs.length - n} others`;
  }

  return new Intl.ListFormat(locale, { style: "long", type: "conjunction" }).format(xs);
}
```
Do you think it makes more sense to move this logic into the project via a utility function, or update it in your `hurdak` library method?

staab:
Hey Zack, sorry I missed this comment, I think moving this into the project itself would be best. That way i18n stuff isn't leaking into the utility library, which increases its scope significantly.

PR #439, opened 2024-10-15
388 - Update the displayList implementation and move into the project

zmjohnso:
This PR addresses #388. Pulled `displayList` into the project and made use of `Intl.ListFormt` instead of pure string concatenation. Maybe this is a slight improvement to the hurdak implementation? Added unit tests, so hopefully the original functionality is maintained. Also `tsconfig` was not happy, saying `Intl.ListFormat` did not exist, so I pointed the tsconfig target to `es2021` so we could have access to it. Not sure about the broader implications of this change.


staab:
Thanks! Looks great.

Issue #349, opened 2024-05-23
Upload images without blocking compose


staab:
Do it like github does

peileppe:
Hello, 

Just discovered coracle .social and it surpasses Primal (or Nostter) in many ways 
Unfortunately Coracle lacks the very simple feature of pasting images into notes
Could this be implemented?

nostr:nevent1qyfhwumn8ghj7ur4wfcxcetsv9njuetn9uq3wamnwvaz7tmjv4kxz7fwwpexjmtpdshxuet59uqsuamnwvaz7tmwdaejumr0dshsz9rhwden5te0d45hxumtv4ujucmvda6kgtcpr9mhxue69uhhqunfweshgefwwfjkgtn8vghxuet59uq3qamnwvaz7tmwdaehgu3wd4hk6tcpz3mhxue69uhhyetvv9ukzcnvv5hx7un89uq3vamnwvaz7tmjv4kxz7fwd4hhxarj9ec82c30qy2hwumn8ghj7un9d3shjtnwdaehgu3wvfnj7qg4waehxw309ahx7um5wghx77r5wghxgetk9uqzqdqsv8tzpjxv3f4xat93jmtzsdryk9prlrvjeec8n2wjrqln6ul2krlmfg

staab:
Yeah, I'll look into it

ticruz38:
#445

PR #442, opened 2024-10-15
inline file upload + dropzone

ticruz38:
This PR address issue #349 

The goal is to have an upload behaviour closed to what is used on Github to avoid blocking the writing workflow while uploading a file

- use the text area as a drop zone for files (only for NoteCreate.svelte atm)
- use the text area to give user feedback, ie [filename is being uploaded using nostr.build]
- the file upload modal has been removed.

ImageInput.svelte now takes an optional `compose` or `textarea` props to update the text content while uploading to provide user feedback.
There is a svelte action in src/util/dropzone.ts that can be attached to any element to support file drop and upload.


ticruz38:
duplicate for nostr-editor

Issue #394, opened 2024-07-24
Restyling user profile page - relays


dtonon:
Related to the new profile (#390), this is the relay section.
I would let the user see the relay info (NIP-11) inline.

![image](https://github.com/user-attachments/assets/e91fc6a6-8773-4932-b197-7b67325dcf18)

![image](https://github.com/user-attachments/assets/e54b8329-f644-4655-a3e9-e558af5a9f37)

ticruz38:
what is the explore button for? send me to the relay marketplace?

staab:
Currently clicking on a relay card opens a special purpose relay feed, we should redirect to the main feed page instead though, with a relay-specific feed loaded.

ticruz38:
> Currently clicking on a relay card opens a special purpose relay feed
When I click on the Relay URL or name, it does open a modal with 2 tabs, notes and review.

So clicking explore would send me to the main feed with notes from this relay only?

staab:
> So clicking explore would send me to the main feed with notes from this relay only?

That's what I was thinking, but on reviewing the code it's not trivial. Just send them to the current relay detail page.

Issue #393, opened 2024-07-23
Restyling user profile page - following & followers


dtonon:
Related to the new profile (#390), these are the following & followers views.
The latter seems a good place to explain why the followers count is tricky on Nostr. 

![image](https://github.com/user-attachments/assets/57a0f390-176b-48b0-a465-8b4923b513a4)

![image](https://github.com/user-attachments/assets/44b5b251-426a-4738-b3e3-7b2d54dc276c)

Issue #390, opened 2024-07-22
Restyling user profile page


dtonon:
This proposal aims to restyle the user profile page with the goal to make the `npub` and the `follow`/`message` CTAs more prominent. The NIP-05 visualization follow the rules defined in #384.
Following and followers are integrated in the horizontal menu, and should be listed on the same view, avoiding opening a different interface (the bottom-up sheet).

![image](https://github.com/user-attachments/assets/5e33c06d-1faa-40f2-9b8f-aa5ac87a243b)

For followed profiles, the "Following" button is hidden and the "Unfollow" action is present just in the top-right menu, because it is a low frequency action.

![image](https://github.com/user-attachments/assets/c75fd709-74ca-46b5-97ec-80bf22c0ac78)

staab:
Beautiful, far superior.

> For followed profiles, the "Following" button is hidden and the "Unfollow" action is present just in the top-right menu, because it is a low frequency action.

I don't know about this. What happens when they click follow? Does the button disappear, or change to unfollow until they re-visit the page? I would find it frustrating to not be able to find the undo button, and it seems unnecessary to hide it.

dtonon:
> I don't know about this. What happens when they click follow? 

Good question. I was thinking of fading out the Follow button and moving up the Message one. Keeping the Unfollow button until the next visit is worse, because the subsequent absence can be disorienting, the user should learn that Unfollow is always inside the "more action" dropdown.
Instead, leaving the Unfollow always visible make it too prominent, and prone to a misclick if one wants to click on Message.
I also thought about the need to undo, but I suppose it is a very infrequent case, the user should know how to solve the problem.

dtonon:
I forgot the user's own profile!
An "Edit" button replace Follows/Message, and the top-right dropdown actions are hidden.

![image](https://github.com/user-attachments/assets/036393a6-ca4c-44d5-a1fb-f5aaeef9cb34)

PR #443, opened 2024-10-16
Profile page restyling

ticruz38:
Related to #390 and #393

Following and followers count are now part of the tabs with a dedicated feed under the profile page information.

#393 is not totally complete, the card still has the "about" for each followers/following profile, I thought it was nice to have it and just waited for your confirmation to actually remove it.

#394 with the relay feed redesign is not part of this PR yet.

Screenshot 2024-10-16 at 16 39 56
Screenshot 2024-10-16 at 16 39 50
Screenshot 2024-10-16 at 16 39 41
Screenshot 2024-10-16 at 16 39 33


src/app/shared/PersonSummary.svelte
```diff
@@ -28,9 +36,17 @@
     </Anchor>
     {#if !hideActions}
       <div class="flex items-start justify-end">
-        <slot name="actions" {pubkey}>
-          <PersonActions {pubkey} />
-        </slot>
+        <div class="flex items-center justify-end gap-2">
+          {#if $following}
+            <Anchor button low class="border-none bg-tinted-800-d" on:click={e => unfollow(pubkey)}
```
staab, line 41: This should use `on:click={unfollowPerson}`

src/app/shared/PersonSummary.svelte
```diff
@@ -28,9 +36,17 @@
...
+              >Followed</Anchor>
+          {:else}
+            <Anchor button accent on:click={e => follow(tagPubkey(pubkey))}>Follow</Anchor>
```
staab, line 44: This should use `on:click={doFollow}`

src/app/views/PersonDetail.svelte
```diff
@@ -57,11 +84,32 @@
...
     .toString()
 
+  $: tabs = [
+    "notes",
+    "likes",
+    "collections",
+    "relays",
+    "following -" + $followsCount,
+    "followers " + $followersCount + "+",
+  ].filter(identity)
```
staab, line 94: Instead of interpolating the strings, use the `tab` slot in the `Tabs` component to customize the display. See Notifications.svelte for an example. You could also use the same badge style.

src/app/views/PersonDetail.svelte
```diff
@@ -73,44 +121,63 @@
...
-      <div class="flex w-full items-center justify-between gap-4">
-        <PersonName class="w-full text-2xl" {pubkey} displayNpubCopyButton />
-        <div class="hidden xs:block">
-          <PersonActions {pubkey} />
+<div>
+  <div class="relative flex gap-4 bg-tinted-700 p-6 text-neutral-100">
+    <div>
+      <PersonCircle {pubkey} class="mt-1 h-12 w-12 sm:h-32 sm:w-32" />
+      {#each actions as action (action.label)}
+        <div animate:flip={{delay: 200, duration: 1000, easing: bounceInOut}}>
```
staab, line 129: This animation is really funky. Maybe use `transition:slide` instead. Also, there's no need to iterate over actions here, just use conditionals, the extra complexity isn't worth it.

https://github.com/user-attachments/assets/6f6a78f9-91be-4689-85e6-e8717386ddd9

staab (changes requested):
Thanks! Some things I noticed:

- [ ] NIP05's have opacity applied, I think they'd look better without opacity. Either daniele's icon or an icon to the left of the nip05 would also be nice.
- [ ] Line 167 of PersonDetail has an empty div that's messing up spacing
- [ ] There should be a bullet or something between nip05 and npub on card headers
- [ ] Now that we're not showing the wot number, let's put the wot arc in line with the username, since it's getting pushed way over to the right by the npub
- [ ] Let's keep follow/unfollow next to "message", daniele's objections notwithstanding
- [ ] Message/follow needs to be replaced by "edit" on the user's own profile page
- [ ] Font size on the user name could be a little bigger
- [ ] Profiles often get opened in modals, maybe use AltColor to make the top part still look like a card
- [ ] Follow/follower counts should be faded and separated by a bullet as in the mockup.
- [ ] The "message" action in the overflow menu can be removed since it's on the main card area already.
- [ ] The wot arc seems to be missing on the profile detail page

staab:
Also, I like keeping the about on the profile cards, good choice.

src/app/views/PersonDetail.svelte
ticruz38: yes the code is clunky, the only reason to use a keyed each here is to have button message gently take the empty space from the disappearing follow button by sliding in.
Not sure I'll get this same behaviour with transition:slide

ticruz38: it's true that the animation does not look that good anyway

ticruz38:
- [x] Removed opacity and added the fa-check icons (could not find Daniele's icon)
- [x] Line 167 of PersonDetail is removed
- [x] Added the bullet character with opacity-75 between nip05 and npub on card headers
- [x] Wot arch is inline with user name
- [x] follow/unfollow button next to "message"
- [x] Message/follow replaced by "edit" on the user's own profile page, Edit action removed from the dropdown as well
- [x] Font size is now text-xl
- [x] ProfileDetails is now wrapped in the Altcolor.svelte utility component, like notes
- [x] Follow/follower counts are now displayed in a Badge, like notifications tabs.
- [x] The "message" action removed from the overflow menu
- [x] Wot arc back next to the user name, scaled up to match the name font-size

staab:
Looks good, I made a few tweaks of my own and pushed to dev. Just a couple final comments.

src/app/shared/PersonActions.svelte
```diff
@@ -22,26 +18,12 @@
...
-  const share = () =>
-    router
-      .at("qrcode")
-      .of(
-        toNostrURI(
-          nip19.nprofileEncode({pubkey, relays: ctx.app.router.FromPubkeys([pubkey]).getUrls()}),
-        ),
-      )
-      .open()
-
```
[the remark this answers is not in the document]
ticruz38: it should be aligned now

staab:
Cool, mind rebasing so I can review?

src/app/shared/PersonName.svelte
```diff
@@ -60,11 +60,10 @@
...
       </div>
-      <div class="flex flex-row items-center gap-1 text-xs">
-        <PersonHandle class="whitespace-nowrap text-accent hidden xs:block" {pubkey}>
-          <span class="opacity-50 hidden xs:block">-</span>
-        </PersonHandle>
-        <p class="whitespace-nowrap opacity-50">{npubDisplay}</p>
+      <div class="flex flex-row items-center gap-2">
+        <PersonHandle class="whitespace-nowrap text-xs text-accent" {pubkey} />
+        <span class="text-xs opacity-75">•</span>
+        <small class="whitespace-nowrap">{npubDisplay}</small>
```
staab, line 66: I made this change because if there's no handle the bullet is just floating out in space, you'll want to adopt my version here.

src/app/views/PersonDetail.svelte
```diff
@@ -197,6 +197,18 @@
...
+      </div>
+      {#if $profile?.website}
+        <Anchor
+          external
+          class="flex items-center gap-2 overflow-hidden overflow-ellipsis whitespace-nowrap"
+          href={ensureProto($profile.website)}>
+          <i class="fa fa-link text-accent" />
+          {stripProtocol($profile.website)}
+        </Anchor>
+      {/if}
```
staab, line 211: This isn't right, I moved website elsewhere, and personabout is now duplicated.

staab (changes requested):
> Merge remote-tracking branch 'upstream/dev' into feat/user-profile

Merge is fine, but rebase is preferred, for future reference.

src/app/shared/PersonName.svelte
ticruz38: messed up the merge, probably a good argument in favor of rebase here

src/app/views/PersonDetail.svelte
ticruz38: I went through your commits, if I'm not mistaking, website is now inline at the tail end of PersonHandle and zapDisplay.

It will go hidden if PersonHandle and zapDisplay are too large, it is hidden on my profile for example.

Screenshot 2024-10-25 at 11 15 02

The row is 80% of the container width atm, I can make it full size and add the flex-wrap condition on the container.
It will push the website to the next line.

Screenshot 2024-10-25 at 11 22 54

Let me know your preference and I'll push the commit eventually

PR #445, opened 2024-10-18
Integrate Nostr-Editor

ticruz38:
Have the note creation to use nostr-editor, a set of nostr extension on top of the tiptap editor.

Bugs and known issue remaining:
- The images are loaded and displayed inside the editor, as a result, the Compose component replace the ImageInput component. The goal is still to keep the underline carrousel, I miss a proper callback to get the uploaded image urls (onComplete not triggering)
- When hitting Enter to select a mention, the editor does break the line, preventDefault does not help
- tags or hashtags are immediately wrapped in <a> tags, it is underlined as you type before "space" key is hit
- nostr-editor comes with a nostr event parsing. Images url are immediately set in the content and in the tags. We could potentially use this functionality entirely and remove internal logic from Coracle, but it depends how confident you are with this dependency.
- Image urls are not in the tags when sending the event, this will be resolved


src/app/shared/Compose.svelte
```diff
@@ -1,26 +1,105 @@
...
+
+      element: element,
+      editorProps: {
+        attributes: {
+          placeholder,
+        },
+      },
+
+      extensions: [
+        StarterKit,
```
staab, line 65: StarterKit has lots of stuff we don't want to support. See flotilla's set up for something much closer to reality.

src/app/shared/Compose.svelte
```diff
@@ -124,27 +151,38 @@
...
+    if (["Enter"].includes(e.code) && suggestions.get()) {
+      e.preventDefault()
+      const pubkey = suggestions.get()
+      if (pubkey) {
+        const content = $editor.getHTML()
+        // only replace the last occurence
+        const replacedContent = content.replace(/(.*)@\w+/, "$1")
+        $editor.commands.setContent(replacedContent, true)
+        $editor.commands.insertNProfile({nprofile: pubkeyEncoder.encode(pubkey)})
+        suggestions.setData([])
```
staab, line 163: This approach will probably not work, tiptap is too magical, you have to do things their way.

src/app/shared/NoteContentTopic.svelte
```diff
@@ -1,16 +1,30 @@
...
+  <NodeViewWrapper data-type="tag" as="span">
+    <Anchor modal underline on:click={onClick}>
+      #{value}
+    </Anchor>
+  </NodeViewWrapper>
+{:else}
+  <Anchor modal underline on:click={onClick}>
+    #{value}
+  </Anchor>
+{/if}
```
staab, line 30: Don't overload NoteContentTopic and PersonLink, just create new components specifically for the editor. Again, see flotilla for how I've structured it there.

src/app/shared/PersonLink.svelte
```diff
@@ -1,16 +1,32 @@
...
+  export let pubkey: string
   export let underline = true
 
-  const path = router.at("people").of(pubkey).toString()
-  const display = deriveProfileDisplay(pubkey)
+  $: attrs = node?.attrs as NProfileAttributes
+
+  $: pubkey = pubkey || attrs?.pubkey || ""
+  $: path = router.at("people").of(pubkey).toString()
+  $: display = deriveProfileDisplay(pubkey)
```
staab, line 18: Avoid too many reactive declarations, the more we can assign with `const` the better for performance. In this case, path will never change if pubkey doesn't, and `deriveProfileDisplay` is already reactive. Unless pubkey changes here, in which case you're doing the right thing.

staab (commented):
> The images are loaded and displayed inside the editor, as a result, the Compose component replace the ImageInput component. The goal is still to keep the underline carrousel, I miss a proper callback to get the uploaded image urls (onComplete not triggering)

I'm not sure why onComplete wouldn't trigger, you may have to debug that in nostr-editor. Let me know if you'd like me to take a look. I do think we can potentially get rid of the image carousel at the bottom and put everything inline. I'd like the editor to show pretty much the same thing as would ultimately be rendered.

> When hitting Enter to select a mention, the editor does break the line, preventDefault does not help

This is probably due to the use of onKeyDown in a way that's not integrated with tiptap. Take a look at how flotilla does mention suggestions via a tiptap plugin, you'll probably have to do something similar.

> tags or hashtags are immediately wrapped in tags, it is underlined as you type before "space" key is hit

I haven't been able to figure out topics. I was able to avoid solving this in flotilla, but it does need to be done. Maybe ignore nostr-editors TagExtension and add a new Node type.

> nostr-editor comes with a nostr event parsing. Images url are immediately set in the content and in the tags. We could potentially use this functionality entirely and remove internal logic from Coracle, but it depends how confident you are with this dependency.

Yep, rip out as much of coracle's code for this and start from scratch. It's outlived its usefulness, although we'll have to make sure the (many) bugfixes make it into the new version. I don't mind the dependency on tiptap, it's what's necessary to bring the UX to the next level.

Cesar and I had also talked about using tiptap to render notes. He was apparently able to get that working, but I've had some trouble. Anyway, I'm not sure it would be performant in practice so we probably shouldn't do it right away. Something to keep in mind though.

As I said before, feel free to nuke NoteImages, Content, NoteReply, NoteCreate, and NoteCreateInline (etc) and come up with new abstractions. I think it's going to be better to tackle the problem completely from scratch.

src/app/shared/PersonLink.svelte
ticruz38: had to change from constant to display the profile, attrs?.pubkey is undefined on the first render

ticruz38:
Took a lot from Flotilla in this commit.

Created a new folder "editor" under the app directory, like Flotilla, but with different Edit component. Let me know if you would rather have it somewhere else, or just add the Edit[whatever].svelte files in the shared directory.

‼️NostrCreateInline and NostrReply still need to be adapted, the code break for any other url than notes/create. I still need to figure out the correct way to submit. For now it's weird, I upload the files and onSubmit() trigger in the onComplete callback of the file upload (kind of unrelated).

- Any view that uses an editor will instantiate Editor with the correct parameters (essentially the session pubkey and eventual mentions)
- Compose.svelte is still there atm, mostly for CSS purpose on the content of the editor (image blur-in blur-out while loaded)
- Have to modify the uploadFile extension to store the files url being uploaded (to put them in the event tags)
- Considering adding a small extension that will count words and characters being typed.
- If we want to be close to a "live" note, I will support Spotify/Youtube/Github link with thumbnail eventually, this will not be a nostr extension but a slightly modified Media component

staab:
Great, looking good.

> For now it's weird, I upload the files and onSubmit() trigger in the onComplete callback of the file upload (kind of unrelated).

Yeah, I think the reason it's this way is so that submit is deferred until files get uploaded. As a result, we currently have to defer uploading files until the user submits the form (which is a feature in a way, because it avoids them uploading files they don't ultimately want to use). I do think that uploading files concurrently for a quick submit is a better way to do this though, so hopefully we can find a way around this pattern.

ticruz38:
- Adapted NoteCreate, NoteReply, NoteCreateInline to use the new Compose element.
- uploading all files before submitting at the moment. editor.commands.uploadFile re-upload all files in the editor atm. ImmediateUpload could be used too, but it does not trigger onComplete callback. 
- added iframes preview for YouTube and Twitter, the twitter preview is being blocked by plausible.io, you might be able to configure it there. If you validate, I'll add those preview to Media.svelte, so we get "native" preview for Tidal, Spotify, Youtube and Twitter. 
- WordCount extension to count characters and words.


Issues:
- When opening/closing a reply, ie creating a draft, the editor will just render everything as text, haven't found how to force parse the content. Let me know if you know..

staab:
> Added iframes preview for YouTube and Twitter, the twitter preview is being blocked by plausible.io, you might be able to configure it there. If you validate, I'll add those preview to Media.svelte, so we'll get a "native" preview for Tidal, Spotify, Youtube and Twitter.

That might be the content security policy, see `frame-src` in index.html.

> When opening/closing a reply, ie creating a draft, the editor will just render everything as text, haven't found how to force parse the content. Do you know how to force it?

I've struggled with this as well. @cesardeazevedo, do you happen to know the solution to this?

> What was the imeta tag used for and what are the parameters to it?

See here: https://github.com/nostr-protocol/nips/blob/07de7ea7e5d2934afac1eb86db0cd0a0200039a0/92.md?plain=1#L7

cesardeazevedo:
> I haven't been able to figure out topics. I was able to avoid solving this in flotilla, but it does need to be done. Maybe ignore nostr-editors TagExtension and add a new Node type.

Not so sure about the actual issue here but the tag extension was very buggy before version 0.0.3.


> Cesar and I had also talked about using tiptap to render notes. He was apparently able to get that working, but I've had some trouble.

I feel like the editor still struggles with things like hardbreak lines, I have found difficult to understand the correct approach.

> I'm not sure it would be performant in practice

I would imagine the nostr-editor to be considerably slower than `welshman/content` (although would love to see some benchmarks), so use it with caution.


> Have to modify the uploadFile extension to store the files url being uploaded (to put them in the event tags)

Nostr-editor indeed needs to provide an easy way to access the uploaded files, gonna work on that.

> I miss a proper callback to get the uploaded image urls

This part indeed needs some improvement.


> When opening/closing a reply, ie creating a draft, the editor will just render everything as text, haven't found how to force parse the content. Do you know how to force it?

Not so sure if I understood correctly, but you can't directly set a event content when initializing the editor

```ts
  const editor = useEditor({
    content: 'Hello nostr:nprofile1qy88wumn8ghj7mn0wvhxcmmv9uq32amnwvaz7tmjv4kxz7fwv3sk6atn9e5k7tcprfmhxue69uhhyetvv9ujuem9w3skccne9e3k7mf0wccsqgxxvqas78x0a339m8qgkaf7fam5atmarne8dy3rzfd4l4x6w2qpncmfs8zh',
  })
```

This will render as text as tiptap won't apply any pasteRules here, I tried to find a way many times but couldn't, you have to call `setEventContent` manually.

```ts
  onCreate({ editor }) {
    editor.commands.setEventContent({
      id: '...',
      content: 'Hello nostr:nprofile1qy88wumn8ghj7mn0wvhxcmmv9uq32amnwvaz7tmjv4kxz7fwv3sk6atn9e5k7tcprfmhxue69uhhyetvv9ujuem9w3skccne9e3k7mf0wccsqgxxvqas78x0a339m8qgkaf7fam5atmarne8dy3rzfd4l4x6w2qpncmfs8zh',
    })
  }
```

Interestingly you can set a HTML representation of the editor state, which is the content you get when copying the content directly from the editor (check the clipboard [inspector](https://evercoder.github.io/clipboard-inspector/))

```ts
  const editor = useEditor({
    content: `<meta charset='utf-8'><p data-pm-slice="1 1 []"><span nprofile="nostr:nprofile1qy88wumn8ghj7mn0wvhxcmmv9uq32amnwvaz7tmjv4kxz7fwv3sk6atn9e5k7tcprfmhxue69uhhyetvv9ujuem9w3skccne9e3k7mf0wccsqgxxvqas78x0a339m8qgkaf7fam5atmarne8dy3rzfd4l4x6w2qpncmfs8zh" pubkey="c6603b0f1ccfec625d9c08b753e4f774eaf7d1cf2769223125b5fd4da728019e" relays="wss://nos.lol/,wss://relay.damus.io/,wss://relay.getalby.com/v1" data-type="nprofile">@</span> </p>`
  })
```

ticruz38:
> I've struggled with this as well. @cesardeazevedo, do you happen to know the solution to this?

It seems to be solved now by instantiating the editor with the content as HTML.

>That might be the content security policy, see frame-src in index.html.

Have finally managed to have tweet preview working while keeping the content security policy, that's 2 twitter script tag, once being labelled with a git tag, but it works

- Had to replicate the FileUpload extension, impossible to get the desired behaviour just by configuring it properly or via the callbacks provided. Might be worth a PR on Nostr-Editor.  Just added another command.getMetaTags()  that will give me the nip-94 image tags present in the note, as provided by the endpoint response. There was also an issue where the updated attributes (src, uploading etc) were placed on the wrong node,  this might be related to 2 extensions manipulating things at the same time.
- I still want to test it with a blossom endpoint, do you know any?
- The new editor has been replicated everywhere compose was used before, should we use it in channel as well?
- Twitter and Youtube preview are now in the Media component, it shows up in the note feed basically
- Images load immediately now, there is a blur-in blur-out animation while it loads. you might upload some image accidentally but submit is much faster.

Happy to get your feedback on the editor, until then I'll just keep merging dev and move on to other issues

ticruz38:
thanks @cesardeazevedo for your answer !

I managed to have it all working well with file upload, I will keep testing and issue a PR on nostr-editor if needed.

I did have that strange issue where file attributes where not reported correctly when using immediate uploads, the examples on nostr-editor repo works as expected, but with the Coracle config, it consistently updated the wrong node, the one right before the actual Image node (pos - 1).

staab:
> Might be worth a PR on Nostr-Editor.

Please do, it's worth having upstream.

> I still want to test it with a blossom endpoint, do you know any?

Try https://blossom.hzrd149.com/

> should we use it in channel as well?

Yeah, let's do it.

--

I'll test the PR, can you rebase from dev to fix the conflicts?

index.html
```diff
@@ -43,6 +44,19 @@
...
+    <script
+      nonce="A132076C"
+      async
+      src="https://platform.twitter.com/widgets.js"
+      charset="utf-8"></script>
+    <script
+      nonce="A132076C"
+      async
+      src="https://platform.twitter.com/js/tweet.d7aeb21a88e025d2ea5f5431a103f586.js"
+      charset="utf-8"></script>
```
staab, line 56: I don't think we should do this, for privacy reasons. It basically makes it possible for twitter to track coracle users, even if the user has turned off link previews/embeds. Can we make this work by just adding it to the csp?

staab (commented):
Sorry to make you undo a fair amount of work, but I don't think it's necessary to render everything as blocks. WYSIWYG is nice in theory, but since we're using a different renderer anyway, we're going to have differences between compose/render. Take a look at how flotilla handles things like links and entities. Pretty much everything should be inline in the editor, and not super rich.

Also, compare the behavior of enter between production and this PR, it's important that enter/shift+enter/meta+enter stay the same. I was testing it out and published a dumb note accidentally because of the change. Try to get the design looking the same too, in terms of colors, textarea height, etc. A lot of finessing has gone into what we currently have.

index.html
ticruz38: Twitter only works this way, no iframe to pick the preview from another origin, it builds it on the fly via the widget library. That's ok the twitter preview from dufflepud is good enough.
Also removed Youtube, the client is blocking a bunch of request despite adding the origin in the frame-src, it seems like the iframe is doing a bunch of subsequent request that are all blocked from the origin. I sometimes get the proper preview but not always

staab:
> My understanding is that mod-enter is to submit, enter and shift-enter is for hardbreak. It is now the default configuration for the editor.

It is in a chat context, but it's actually the reverse when composing a note. Sort of weird, but seems the most intuitive.

staab:
Something is very wrong with some of these PRs, try checking dev out and cherry-picking your changes, then doing `git rebase -i dev`. There are a whole bunch of changes from dev that are being merged back in here. When reconciling branches, try to always `rebase` rather than `merge`.

ticruz38:
> There are a whole bunch of changes from dev that are being merged back in here. When reconciling branches, try to always rebase rather than merge.

I rebased from origin/dev indeed, I had to go through twenty-ish commits cherrypicking the right changes, I might be doing something wrong, I'm not very used to rebase
Do you mean this PR is reversing some changes from dev?

staab:
It's the reverse, the diff is including changes that already exist on dev. This is likely due to me force-pushing to dev at some point in the past after a merge (which is why rebase works better). Try rebasing with `-i` and change `pick` to `squash` for all listed commits. That will give you a single commit to work with. Or check out a fresh copy of dev, then `git reset feat/nostr-editor` to pull your changes over without any history.

staab:
I'm happy to hop on a call and work on it with you if that would be helpful

src/app.css
```diff
@@ -236,8 +236,35 @@ body,
...
+  /* @apply bg-primary text-primary-content; */
 }
+
+.tiptap p.is-editor-empty:first-child::before {
+  color: #adb5bd;
+  content: attr(data-placeholder);
+  float: left;
+  height: 0;
+  pointer-events: none;
+}
```
staab, line 269: Nice! I will steal this for flotilla.

src/app/editor/FileUpload.ts
```diff
@@ -0,0 +1,396 @@
...
+    const tasks = this.findNodes(false).map(([node]) => {
+      return this.upload(node)
+    })
+    try {
+      this.editor.storage.fileUpload.loading.set(true)
+      await Promise.all(tasks)
+      this.options.onComplete(this.editor)
+    } catch (error) {
+      console.error(error)
+      this.editor.storage.fileUpload.loading.set(false)
```
staab, line 267: No need for this bit, finally will still run if an error happens

src/app/shared/NoteCreateInline.svelte
```diff
@@ -48,26 +53,24 @@
...
 
-    const content = compose.parse().trim()
+    const content = editor.getText().trim()
 
     if (!content) return showWarning("Please provide a description.")
 
     if (!skipNsecWarning && content.match(/\bnsec1.+/)) return nsecWarning.set(true)
 
-    const tags = [...tagsFromContent(content), ...getClientTags()]
+    const tags = [...tagsFromContent(content), ...getClientTags(), ...editor.commands.getMetaTags()]
```
staab, line 66: Seems like `tagsFromContent` could be replaced by nostr-editor's new tags functionality, or at least `getEditorTags`

src/app/shared/NoteCreateInline.svelte
```diff
@@ -82,25 +85,39 @@
...
+      <!-- placeholder="What's up?" -->
+      <Compose bind:element={editorElement} {editor} style="min-height: 3em;" />
       <div class="flex items-center justify-between">
         <div class="flex items-center justify-end gap-3">
           <i class="fa fa-cog cursor-pointer" on:click={() => options.setView("settings")} />
-          <ImageInput multi hostLimit={3} on:change={e => images.addImage(e.detail)}>
-            <i slot="button" class="fa fa-paperclip cursor-pointer" />
-          </ImageInput>
+          <button
+            class="hover:bg-white-l staatliches flex h-7 w-7 cursor-pointer items-center justify-center gap-2 whitespace-nowrap rounded bg-white px-6 text-xl text-black transition-all"
```
staab, line 117: This button should look like the others. Also, be careful about using `black` and `white`, which won't play well with light theme.

staab:
Some display bugs and behavior issues:

- Ctrl+enter doesn't submit the form on reply, but enter does. It should be the reverse
- Nothing should render as a block, currently invoices, quotes, links do
- link-content doesn't look as good as it does in flotilla. It should have some kind of background to it.
- There are some weird artifacts with the mention suggestion dialog. Also, the size changes and the arrow moves, they should stay still.
- 

Screenshot 2024-10-31 at 8 47 22 AM
Screenshot 2024-10-31 at 9 08 15 AM

Also, the PR needs a rebase. I ran one locally and it worked fine, so I think the reset did the trick.

src/app/shared/NoteCreateInline.svelte
ticruz38: Cesar recently started working on it, it probably isn't ready yet.
getEditorTags will gives you all #whatever in the editor. It refers to hashtags not nostr event tags

ticruz38:
* reply now submit on mod+enter and not enter
* Editor link, event, address and invoices render like in flotilla except the background is a different colour
* The suggestions display used to be like flotilla (animation and so on), now it looks closer to Coracle, a lighter background for selected item, no animation.

staab:
> reply now submit on mod+enter and not enter

The main note create should work the same way too, right now enter sends the note.

The popover looks better, but there's still this:

Screenshot 2024-11-04 at 11 04 30 AM

For blossom, try https://cdn.satellite.earth/, that one should work.

ticruz38:
> For blossom, try https://cdn.satellite.earth/, that one should work.
This one work, but you need to zap the server, the data is then kept for a month. This enpoint requires an additional `size` tag that is not required in BUD02 specs, strange.

I have added the possibility to switch from nip96 to blossom in user settings. 

Most blossom server right now seems to be missing the any domain cors settings, The only one that is accessible from anywhere so far is satellite.

staab:
Looks pretty good, although it needs a rebase, and I'm getting some errors about `getMetaTags` not existing. A few other things:

- The paperclip button should look like the other buttons
- Use this for `link-content`:

```
.link-content,
.tiptap [tag] {
  @apply max-w-full overflow-hidden text-ellipsis whitespace-nowrap rounded bg-neutral-200 px-1 no-underline;
}

.link-content.link-content-selected {
  @apply bg-neutral-300;
}
```

- [nit] I have a particular order to imports: external library (svelte, ramda, hurdak), welshman, lib/partials, app. Just try to keep stuff in that order, it makes it easier to look for things. I know this is weird.

staab (commented):
Getting there!

- [ ] There is a difference in newline rendering between edit and render (see image below). This is probably just a matter of using `this.editor.getText({blockSeparator: "\n"})` when calling getText.
- [ ] https://github.com/coracle-social/coracle/pull/445/files#r1824731976
- [ ] For some reason, mentions don't work right after topics (see video below)

Screenshot 2024-11-07 at 1 41 29 PM

https://github.com/user-attachments/assets/ae81d487-2736-4329-8272-11faf0f359ec

I also ran into this error when pasting a `nevent`:

```
proxy.js:385 Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at Array.displayEvent (EditEvent.svelte:16:20)
    at Array.create_default_slot (EditEvent.svelte:36:18)
    at create_slot (utils.js:165:22)
    at create_dynamic_element (EditorContent.svelte:34:24)
    at create_fragment2 (NodeViewWrapper.svelte:13:11)
    at init (Component.js:148:34)
    at new NodeViewWrapper (NodeViewWrapper.svelte:14:21)
    at create_fragment (EditEvent.svelte:35:65)
    at init (Component.js:148:34)
    at new EditEvent (EditEvent.svelte:27:14)
```

Pasting npubs is also buggy:

https://github.com/user-attachments/assets/eea297c4-2f03-469b-b742-0b4e8919f4e2

Background for link-content looks bad in onboarding (as well as in replies and chat):

Screenshot 2024-11-07 at 2 09 50 PM

src/app/editor/FileUpload.ts
```diff
@@ -0,0 +1,394 @@
...
+    const url = res.nip94_event?.tags.find(x => x[0] === "url")?.[1] || ""
+    const sha256 = res.nip94_event?.tags.find(x => x[0] === "x")?.[1] || ""
+    return {
+      url,
+      sha256,
+      tags: res.nip94_event?.tags.flatMap(item => item.join(" ")),
+    }
+  } catch (error) {
+    throw new Error(error as string)
+  }
```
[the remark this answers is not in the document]
ticruz38: indeed

ticruz38:
- [X] Newline rendering issue was solved with your suggestion
- [X] Have modified tag so that we can immediately mention people, The behaviour has slightly change,  tag is now display once we hit the space key
- [X] I could not reproduce the issue, however it will no longer break if the content is undefined or can't be resolved.
- [] I tried with different npub with different editor content, I could not reproduce the bug. Can you send me the content or more step on how to reproduce the bug, does it happen every time?
- [X] Background link, mention, tags etc are unchanged, but the text content will now be the "reverse" of the background, so the contrast between text and background is always high.

staab (commented):
Merged 🎉 although I have a couple small changes you can follow up on

src/app/views/UserSettings.svelte
```diff
@@ -14,15 +14,16 @@
   import {fuzzy} from "src/util/misc"
   import Select from "src/partials/Select.svelte"
 
-  const values = {...$userSettings}
+  const values = $userSettings
```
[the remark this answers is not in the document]
ticruz38: I think the $ calling get on the store does give me a local copy already

src/app/editor/index.ts
```diff
@@ -72,7 +72,9 @@ export const getEditorOptions = ({
   submitOnEnter = true,
   submitOnModEnter = false,
   element,
-  defaultUploadUrl = getSetting("nip96_urls").slice(0, 1)[0] || "https://nostr.build",
+  defaultUploadUrl = getSetting("upload_type") == "nip96"
+    ? getSetting("nip96_urls").slice(0, 1)[0] || "https://nostr.build"
+    : getSetting("blossom_urls").slice(0, 1)[0] || "https://cdn.satellite.earth",
```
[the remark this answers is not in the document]
ticruz38: Good question...

PR #456, opened 2024-10-29
Improve onboarding flow

staab:
This PR removes pubkey and private key login entirely, and uses DNS-based remote signers as the recommended option.


ticruz38 (approved):
Have been able to login and signup with an nsec.app domain, it works.Have not been able to try with an nsec bunker yet.

I will configure one in my free time using [this](https://github.com/kind-0/nsecbunkerd)

Issue #369, opened 2024-06-13
Undo


staab:
- [ ] Allow user to set delay in settings
- [ ] Show toast with undo link for top-level notes
- [ ] Use Daniele's mockup for reply undo/progress

Issue #235, opened 2024-01-15
Improve submission status for new responses


dtonon:
Following up #233 (_Replies don't show up immediately_) I propose a new UI to show the posting/posted status of a note.
The note is placed in the correct position on the thread and the bottom bar is replaced with a loading animated component:

<img width="767" alt="Screenshot 2024-01-15 alle 21 16 21" src="https://github.com/coracle-social/coracle/assets/89577423/ac54786b-2dcb-47d8-afc8-05c7ad48742f">
<br/><br/>

After the submission procedure ends the summary is displayed on the same bar, on the left a buttons permit to see the details (current bottom sheet modal):

<img width="768" alt="Screenshot 2024-01-15 alle 21 23 03" src="https://github.com/coracle-social/coracle/assets/89577423/6295a3c0-045f-4731-a4e3-aa81d547c7de">
<br/><br/>

The bar can be replaced after a certain amount of time (ex. 30 seconds) or if the note receives an event (reply, reaction, zap).

dtonon:
We should probably add the case of an unsuccessful submission (e.g. network error) or one with a high percentage of missing relays, giving the user the option of retrying the submission with a click.

ticruz38:
> We should probably add the case of an unsuccessful submission (e.g. network error) or one with a high percentage of missing relays, giving the user the option of retrying the submission with a click.

@hodlbod ? Reopening this issue if the retry is necessary

staab:
I think it's ok for now, we could probably improve the HUD before adding inline retry

PR #457, opened 2024-10-31
Creation and Submission delay and status

ticruz38:
This PR address issues #369 and #235

* New setting to delay the publishing of a note (default is 0)
* Cancelable toast when creating a new note
* Inline Note status when replying to a note.

When replying, there is a (configurable) 5 seconds countdown before the note is sent
Screenshot 2024-10-31 at 16 40 06

Once the note is sent, it tracks the pending requests via a loading bar
Screenshot 2024-10-31 at 16 40 12

Once all request are settled, it shows a status summary, with the possibility to open a more detailed modal
Screenshot 2024-10-31 at 16 40 25

The post status is displayed for 30 seconds, after which the actions buttons shows up like a normal note
Screenshot 2024-10-31 at 16 40 53


staab:
This diff is wonky too. For some reason it's showing all my commits as your commits. I don't know what you're doing to make that happen, but can you reset/rebase to clean it up?

src/app/shared/Note.svelte
```diff
@@ -42,6 +42,8 @@
     sortEventsDesc,
   } from "src/engine"
 
+  import NotePending from "./NotePending.svelte"
```
staab, line 45: Relative imports are evil, use a full import path

src/util/misc.ts
```diff
@@ -1,11 +1,17 @@
...
 import {pluck, fromPairs, last, identity, sum, is} from "ramda"
 import {Storage, ensurePlural, seconds, tryFunc, round} from "hurdak"
 import Fuse from "fuse.js"
 import logger from "src/util/logger"
 
+export const timestamp1: Readable<number> = derived([], (_, set) => {
+  setInterval(() => {
+    set(Math.floor(Date.now() / 1000))
+  }, 1000)
+})
```
staab, line 13: This is going to result in a memory leak, since the interval never gets canceled.

src/app/shared/Note.svelte
```diff
@@ -102,11 +104,21 @@
...
 
-  const addToContext = events => {
-    context = context.concat(events)
+  const addDraftToContext = (event, cb) => {
+    event.status = "draft"
+    event.remove = () => {
+      cb()
+      removeFromContext(event)
+    }
+    context = context.concat(event)
```
staab, line 117: Mutating events isn't a good idea. I don't think it will technically break anything, but all this stuff gets serialized and shoved into indexeddb, which will get weird fast. See if you can find another way to pass this data between components.

staab (commented):
You'll need to rebase this on dev since I merged the groups stuff. Also, take a look at @welshman/app's `thunk` module, it solves a lot of these problems much more cleanly. I'm not sure if we can integrate it cleanly yet, but see if you run into any problems with that.

src/app/shared/Note.svelte
ticruz38: Yeah I tried, but it's difficult because Note is used recursively, having the event state and functionality on the event object simplified a lot of stuffs.
Aren't those event already pulled and serialised from indexedDB?

ticruz38:
> thunk module, it solves a lot of these problems much more cleanly. I'm not sure if we can integrate it cleanly yet, but see if you run into any problems with that.

Thunk is clean for NoteCreate
For the reply, it's feasible too but the thunk will have to travel across 3 or 4 components like this.
NoteReply > Note(parent) > Note(reply) > NotePending
When using the publish command it populates the $publishes store that I can then import directly in NotePending and check out the state of my pub.

staab:
> EDIT: No I need a callback, the thunk status or result only resolved once the relay replied, I need either one more status (SENT?) or a callback.

Look for the `Pending` status, that gets set when the event is first sent to the relay. If `pub.status` is empty, that means the event hasn't been sent yet. You should also be able to pass `delay` to the thunk to handle the timeout.

> For the reply, it's feasible too but the thunk will have to travel across 3 or 4 components like this.
NoteReply > Note(parent) > Note(reply) > NotePending

I'm actually hoping to migrate from publishes to thunks (which are a superset of the functionality), this issue is actually a step in that migration. We can just add these to `engine/state` so they're available from anywhere. If you want to go ahead and do that migration as part of this issue, and get rid of `publishes` entirely, go ahead, they're only used in a couple places.

ticruz38:
ok I'll make the migration to thunk, almost there

ticruz38:
Just finished the thunk migration and rebased from dev

Might be worth another look

staab (commented):
When delay = 0, the tooltip still shows while the remote signer responds:

Screenshot 2024-11-12 at 4 07 58 PM

Not sure why, but a duplicate thunk is getting added with an undefined ID:
Screenshot 2024-11-12 at 4 10 08 PM

.husky/pre-commit
```diff
@@ -1 +1 @@
-npm run check
+
```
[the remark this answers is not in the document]
ticruz38: caught cheating on the pre-commit hook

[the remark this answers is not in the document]
ticruz38: because of the way I link welshman, it is checking welshman as well for some reason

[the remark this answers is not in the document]
staab: You mean you get errors from the welshman project? There shouldn't be any if you're on the latest version, maybe pull? Also, you can bypass pre-commit if needed by doing `git commit -n`

src/engine/state.ts
```diff
@@ -802,27 +801,17 @@ export const publish = async ({forcePlatform = true, ...request}: MyPublishReque
...
-  const pub = basePublish(request)
-
-  // Listen to updates and update our publish queue
-  if (shouldTrackPublish(request.event)) {
-    const pubInfo = omit(["emitter", "result"], pub)
+  const thunk = publishThunk(request)
 
-    pub.emitter.on("*", t => publishes.update(assoc(pubInfo.id, pubInfo)))
-  }
+  thunks.update(assoc(request.event.id, thunk))
```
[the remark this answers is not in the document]
ticruz38: I could not find that store, what's the name of it?

src/app/shared/Note.svelte
```diff
@@ -283,18 +298,22 @@
...
-                bind:this={actions}
-                {removeFromContext}
-                {addToContext}
-                {replyCtrl}
-                {showHidden}
-                {replies}
-                {likes}
-                {zaps}
-                {muted} />
+              {#if !isDraft || event.created_at < $timestamp1 - 45}
```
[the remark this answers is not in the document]
ticruz38: I did test it today, each reply keeps its own state and countdown

src/engine/state.ts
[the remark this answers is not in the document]
staab: Right here: https://github.com/coracle-social/welshman/blob/master/packages/app/src/thunk.ts#L104

src/app/shared/NoteActions.svelte
```diff
@@ -138,7 +133,7 @@
         id: note.id,
         anonymous: Boolean(note.wrap),
       })
-      .cx({callback: addToContext})
+      .cx({callback: repository.publish})
```
staab, line 136: This can actually be removed. We might be able to get rid of `callback` from Zap as well if it's not used anywhere else.

src/partials/ThunkStatus.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+  import type {Thunk} from "@welshman/app"
+  import {PublishStatus} from "@welshman/net"
+  import Anchor from "src/partials/Anchor.svelte"
+
+  export let thunk: Thunk
+
+  const status = thunk.status
+  $: total = thunk.request.relays.length
+  $: pending = Object.values($status).filter(s => s.status === PublishStatus.Pending).length
```
staab, line 10: This needs to remove LOCAL_RELAY_URL from the count

src/app/views/NoteCreate.svelte
```diff
@@ -139,11 +140,28 @@
           ],
         }),
     })
+    const thunk = publish({
+      event: template,
+      relays: ctx.app.router.PublishEvent(template).getUrls(),
```
staab, line 145: This template doesn't have a pubkey yet, so random relays are selected. You'll need to sign the event first.

src/app/views/UserSettings.svelte
```diff
@@ -55,6 +55,14 @@
...
       </p>
     </Field>
+    <Field>
+      <div slot="label" class="flex justify-between">
+        <strong>Send Delay</strong>
+        <div>{values.send_delay} ms</div>
+      </div>
+      <Input type="range" step="500" bind:value={values.send_delay} min={0} max={15_000}></Input>
+      <p slot="info">A delay period allowing you to cancel a reply or note creation, in seconds.</p>
+    </Field>
```
staab, line 65: Let's display seconds instead of ms. A step of 1 is fine

src/app/views/NoteCreate.svelte
staab: Also, when using nip 46, there's a pretty long delay before a note gets published. This isn't really part of this PR, but go ahead and add loading to NoteCreate to avoid double publishes.

src/engine/state.ts
```diff
@@ -151,7 +151,7 @@ export const hasNip44 = derived(signer, $signer => Boolean($signer?.nip44))
 // Base state
 
 export const anonymous = withGetter(writable<AnonymousUserState>({follows: [], relays: []}))
-export const publishes = withGetter(writable<Record<string, PublishInfo>>({}))
+export const groupHints = withGetter(writable<Record<string, string[]>>({}))
```
staab, line 154: groupHints is old code, it can be removed

src/partials/Toast.svelte
```diff
@@ -75,10 +82,16 @@
...
+  const timeInterval = setInterval(() => timeLeft > 0 && timeLeft--, 1000)
+
+  onDestroy(() => {
+    clearInterval(timeInterval)
+  })
 </script>
 
 {#if $toast}
-  {#key "key"}
+  {#key $toast.id}
```
staab, line 94: This was actually correct, it's to avoid an animation when there is already a toast visible. Unless something changed?

ticruz38: I have a weird behaviour with the "key" key where the content of 2 different toast is merged together. 

I have the first delay toast staying for 3 seconds followed by the "publish" toast with a 5 seconds timeout. If I don't have a uniq key, then both toast content are merged together until the 5 seconds timeout of the second toast.

PR #463, opened 2024-11-04
Delete all group related functions

ticruz38:
Continuation of #460 

- Removed Group related projections
- Removed all commands linked to group
- Removed all group models
- Removed all group state variable.
- Removed all partial group components

refactored and renamed publishToZeroOrMoreGroup, it is mostly used as sign and send, but in the case of reply or repost, it might still be necessary to publish to groups, let me know and I'll revert that one change.

Overall this might be too aggressive, do we show group summary in "standard" feed? If that's the case, I'll have to bring back a bunch of state function and group components


staab (commented):
> refactored and renamed publishToZeroOrMoreGroup, it is mostly used as sign and send, but in the case of reply or repost, it might still be necessary to publish to groups, let me know and I'll revert that one change.

This is a good point, I want to completely drop support for groups since I don't think nip 72 or 87 are the way forward (or, at least are irrelevant to coracle for now). But we'll also want to run a migration to delete events posted to encrypted groups so that they don't show up in the app and end up getting replied to. The best way to do this would probably be to update `migrateEvents` to filter out events with `event.wrap.tags.find(t => isGroupAddress(t[1]))`.

> Overall this might be too aggressive, do we show group summary in "standard" feed?

This is probably the one thing we should keep. Instead of linking internally to the group, just link out using NIP 89 handlers/wot to nostrudel/satellite etc.

Other things:

- [x] Be sure to run `npm run format && npm run check`. Feel free to add it to a pre-commit hook via husky or whatnot.
- [ ] Cross-post note action and dialog need to be removed
- [ ] FORCE_GROUP can be removed everywhere
- [ ] Group stuff from `unwrapRepost` can be removed
- [ ] Group stuff from InviteCreate can be removed
- [ ] `contextAddress` from Feed, Note, NoteReply, NoteActions
- [ ] `isContextFeed` from FeedFormSaveAsList
- [ ] `isContextAddress` from util/feeds
- [ ] `isGroup` from NoteContentQuote
- [ ] group stuff in `UserKeys` can be removed
- [ ] `attemptedAddrs` and `getStaleAddrs` in engine/requests, along with comment above
- [ ] `mentionGroup` in engine/state
- [ ] group stuff in onboarding
- [ ] groups in EventEdit
- [ ] `showGroup`
- [ ] groups stuff in app/utils/router
- [ ] empty `loadAppData` function

Also, a follow up to this for welshman: https://github.com/coracle-social/welshman/issues/11

ticruz38:
It should be it with the groups now, have deleted all occurrences.

There is also a pre-commit check now, on format, types and errors

staab:
Awesome, reset/merged.

PR #465, opened 2024-11-07
Fix feeds

staab:
Feeds were getting pretty slow, this release speeds them up again, and simplifies a bunch of stuff. It includes some upgrades to welshman related to feeds.


ticruz38:
Seems indeed faster to me, and I couldn't find any bug.

I never see any repost  in my feed when selecting "follows" as my default filter, is it normal?

staab:
Coracle doesn't fetch reposts unless you specifically ask for them in the feed builder

PR #467, opened 2024-11-11
Nostrconnect

staab:
This adds a nostrconnect:// login workflow as a default, instead of username-based.


ticruz38:
are you sure this to be merged into master?

staab:
Ah, yep good catch.

Issue #453, opened 2024-10-29
Remove calendars


staab:
Also make sure to remove the libraries we use to render calendars.

Issue #452, opened 2024-10-29
Remove listings


staab:
Leave the code for rendering listings in notes.

PR #468, opened 2024-11-12
Remove listings and events

ticruz38:
Removing calendar events as well as the @event-calendar dependency.

EventDetail, EventEdit and EventDelete are still accessible via an Event Note so they remain.

#452
#453


src/app/views/NoteCreate.svelte
```diff
@@ -126,18 +125,6 @@
...
-      calendar_event: () =>
-        createEvent(31923, {
-          content,
-          tags: [
-            ...tags,
-            ["d", uuid()],
-            ["title", opts.title],
-            ["location", opts.location || ""],
-            ["start", dateToSeconds(opts.start).toString()],
-            ["end", dateToSeconds(opts.end).toString()],
```
staab, line 138: These can be removed from other places in the file.

staab (commented):
Go ahead and remove the EventActions stuff, and the delete/edit links from EventInfo. That will allow us to remove even more. And since users can't create events in coracle any more, it doesn't make sense to let them edit/delete.

ticruz38:
- removed all things related to listing and calendar in NoteCreate
- remove EventDetail and EventDelete and all actions/buttons/links referring it

staab:
You should be able to completely remove EventActions and EventDetail. Be sure to remove the `detailPath` stuff in `EventInfo` as well.

ticruz38:
- Removed EventDetail too
- Kept EventAction only for the QR code sharing, if not, I'll remove EventActions completely

staab:
Yeah, go ahead and remove EventActions completely. We can add qr code sharing to normal event share if we decide we want that down the line.

Issue #448, opened 2024-10-24
Make sure we use outbox model when loading naddr


staab:
Regardless of whether the naddr has relay hints, we should load the pubkey and follow the write relays.

naddr1qvzqqqr4xqpzqlcm95syvm74l7pkc672rgyyn7w8w93vlsqt0ecg4y2zwc7qy9nnqqzx6at5v5el5p2k

PR #476, opened 2024-11-18
Use pubkey relay if naddr does not provide relay hints

ticruz38:
Fallback to the pubkey write relays when naddr to not provide relay hints.

In deriveEvents, when loading the event, make sure we have some relay set in the request.
In NoteDetails and EventDetails, make sure we pass in some relays url so that we can load the event in the onMount callback

#448

### DIFF
--- a/src/app/views/Bech32Entity.svelte
+++ b/src/app/views/Bech32Entity.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import {Address} from "@welshman/util"
   import {loadHandle} from "@welshman/app"
+  import {ctx} from "@welshman/lib"
   import Content from "src/partials/Content.svelte"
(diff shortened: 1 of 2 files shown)


src/app/views/Bech32Entity.svelte
```diff
@@ -16,10 +17,14 @@
   <NoteDetail id={data} {relays} />
 {:else if type === "naddr"}
   {@const address = new Address(data.kind, data.pubkey, data.identifier).toString()}
+  {@const relays = [
+    ...(data.relays || []),
+    ...Array.from(ctx.app.router.FromPubkey(data.pubkey).getUrls()),
```
staab, line 22: Array.from shouldn't be necessary any more

staab (approved):

Issue #480, opened 2024-11-20
Refine delayed send


staab:
The note create dialog just sits there while the delay happens, which forces the user to wait, and allows them to double-submit. Close the dialog, and when the user hits cancel, open the dialog back up with the note contents available to edit.

https://github.com/user-attachments/assets/1b75eaab-762c-404d-b722-4e4a400cba1d

If a user cancels sending a note reply, the text of the note is lost. Save the draft if they cancel. Also, the display before sending is kind of weird because it looks un-aligned. Start the orange highlight at the left, reaching maybe 20% across. Also, the animation looks like it's using a bounce interpolation, just make it smooth.

https://github.com/user-attachments/assets/b60b1a06-9289-4f67-bfbe-4bb665451fc6

staab:
One other thing, if there are 0 failures or 0 timeouts, no need to show that extra info:

![Screenshot 2024-11-21 at 1 34 34 PM](https://github.com/user-attachments/assets/ab8e32b2-fab9-4662-9df3-8c62aac85a77)

PR #481, opened 2024-11-21
Refine delayed send

ticruz38:
NoteCreate:
- Close the modal once the event is signed
- Re-open the modal when/if the sent is cancelled
- Keep an entry in the draft store with id "note-create" to keep the note content during the modal lifecycle

NoteReply:
- Save the draft if the send is cancelled
- Align the note content text with the countdown text
- loading start at 20% to the left, the loading bar is now tweened animated rather than spring animated, for a smoother look
- The loading bar has a lighter background now as Daniele's screenshot
- When a Note is being replied, a reply with a background accent appear on the parent note, per Daniele's screenshot.
- The thing missing from Daniele's screenshot is the action popover next to the loading bar, the note actions (deleted, broadcast repost etc) can only be used once the event is actually uploaded to one relay, so it should probably slide in before the upload is complete, let me know if I should add it anyway.

### DIFF
--- a/src/app/shared/Note.svelte
+++ b/src/app/shared/Note.svelte
@@ -32,6 +32,7 @@
   import NoteActions from "src/app/shared/NoteActions.svelte"
   import NoteContent from "src/app/shared/NoteContent.svelte"
   import NotePending from "src/app/shared/NotePending.svelte"
+  import {drafts} from "src/app/state"
   import {router} from "src/app/util/router"
   import {
     env,
@@ -74,7 +75,12 @@
 
   const addDraftToContext = (event, cb) => {
     draftEventId = event.id
-    removeDraft = () => cb() && repository.removeEvent(event.id)
+    removeDraft = () => {
+      cb()
+      repository.removeEvent(event.id)
+      drafts.set(note.id, event.content)
+      replyIsActive = false
+    }
   }
 
   const onClick = e => {
@@ -290,6 +296,7 @@
                   note={event}
                   zapper={$zapper}
                   {replyCtrl}
+                  {replyIsActive}
                   {showHidden}
                   {replies}
                   {likes}
--- a/src/app/shared/Note.svelte
+++ b/src/app/shared/Note.svelte
@@ -78,7 +78,6 @@
     draftEventId = event.id
     removeDraft = () => {
       cb()
-      repository.removeEvent(event.id)
       drafts.set(note.id, event.content)
       sleep(10).then(() => {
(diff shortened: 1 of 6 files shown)


ticruz38:
#480

src/app/shared/Note.svelte
```diff
@@ -74,7 +75,12 @@
 
   const addDraftToContext = (event, cb) => {
     draftEventId = event.id
-    removeDraft = () => cb() && repository.removeEvent(event.id)
+    removeDraft = () => {
+      cb()
+      repository.removeEvent(event.id)
```
staab, line 80: If you're using `publishThunk`, this shouldn't be necessary, welshman does it automatically.

ticruz38:
- I made this change so that the countdown does go to zero, it's a Math.ceil, If I start at 5 the note will send at 1
- the orange reply button is in the @dtonon mockup, not sure how to display/use it, my understanding is to show it when the note is being replied
- damn I thought I had the reply loading bar right this time

ticruz38:
- Reply button is gone
- NoteReply does reopen on cancel

On your last point I'm confused, are you speaking of the loading bar?

Daniele's mockup
Screenshot 2024-11-22 at 18 02 03

App rendering
Screenshot 2024-11-22 at 17 50 26

The gradient goes from var(--accent) to transparent, I don't see an outline in Daniele's design, I might be looking at the wrong thing though

staab:
My bad, I was looking at your first version as the authoritative mockup 🤦‍♂️ What you have now looks fine, the one thing would be to add something to the state pre-publish, which looks pretty naked. I like what you had originally, with the outline matching the button color

src/app/shared/NotePending.svelte
```diff
@@ -62,25 +65,32 @@
...
     {:else}
-      <span>Published to {success}/{total} ({failed} failed, {timeout} timed out)</span>
+      <span
+        >Published to {success}/{total}
+        {#if failed > 0 || timeout > 0}
+          ({failed > 0 ? failed + " failed" : ""}{timeout > 0
+            ? (failed > 0 ? ", " : "") + timeout + " timed out"
+            : ""})
+        {/if}
+      </span>
```
staab, line 85: TERNARY MADNESS

staab (approved):

src/app/shared/NotePending.svelte
ticruz38: 🤯
I agree, I'll make a function in the script

Issue #371, opened 2024-06-18
Show notices to users


staab:
Maybe via hud. If there's an error related to a note they've published, show that more prominently. A good test case is sending nip 72 meta to damus.

ticruz38:
A note that fails to broadcast to any relay is displayed as such:

<img width="288" alt="Screenshot 2024-11-20 at 18 11 59" src="https://github.com/user-attachments/assets/44d12c23-408a-47ba-971f-5d4a17553ea7">

In that case it's a timeout, for all relays. 

Should we use a toast if any specific reason is sent back by all or some relay that would give us a better hint as to why the note failed?

staab:
Ok, here's what I'm thinking. 

-[ ] Leave the HUD alone, but make the dialog a tabbed view, with "Events", "Connections", and "Notices".
- [ ] Events would be the current view.
- [ ] Connections would have a set of stats boxes at the top for the quality of various connections. Basically the same logic as the status dot on relay cards, but for all connections available in `pool`, sorted by connection quality descending.
- [ ] Add a "warnings" badge to RelayCard that displays if there have been notices or refused events. If you click on it, it opens the notices tab filtered down to that relay. Probably create an entirely new relay card, since that component is getting pretty overloaded and will have to be re-worked next release anyhow.
- [ ] On the "notices" tab show a multi-select at the top which allows the user to pick 0 or more relay urls from the pool. Below, show a list of notices for that relay ordered chronologically descending, or "No notices found" if not.

PR #487, opened 2024-11-22
Enhance note status

ticruz38:
Related to #371 

- Adding 3 tabs to the Publishes view
- Connections have 6 stats boxes that display the connection quality (connected, logged out, failed etc) clicking one of the stat box filter out the connections that fit in that category.
- Each connection or Relay has a badge if some event where posted via them (success is also used)
- Notices tab will list all event posted via this relay and the status attached

Notices are kind of unclear what should be displayed, I have the event, the status of the request and the relay url, not sure how to display this. 

This is a draft, just look at the screenshot and give me some general feedback

Stats box are likely too big, you can see a timeout attached to some relay
Screenshot 2024-11-22 at 16 00 30

Here the relay display how many successful events where posted to it
Screenshot 2024-11-22 at 16 37 31

Screenshot 2024-11-22 at 16 01 31

Screenshot 2024-11-22 at 16 38 48


staab:
Nice, a few comments:

- [ ] Instead of icons, use text. In this case, the icon is the main piece of information so it shouldn't be hidden behind something the user has to guess at or hover over. Something similar to the current status dot would be good — maybe in the top right, across from the relay name, show a dot + a status name with a tooltip that explains what the status means.
- [ ] I agree the tiles are too big. Maybe use SelectButton multiple=true instead to control which connections are visible
- [ ] On notices, I was thinking less about showing event publishes than NOTICE or OK=false messages from a given relay. If it's an ok related to an event, you can show some event details, but the main thing to surface is the message that goes along with the notice or ok.

ticruz38:
Show connection status, there is no popover on the status as the name is quite descriptive already.
Screenshot 2024-11-25 at 14 33 57

With a status selected using SelectButton
Screenshot 2024-11-25 at 14 47 42

The notice looks like logs in a terminal, since Notice are de-facto a nerdish info, it make sense to have this more technical oriented display.
Screenshot 2024-11-25 at 14 24 49

I rarely get any message from the endpoint along with the status, I have here the default message being hardcoded as  "Published", I can use something more technical such as 200 OK
When clicking on the line it pops a modal with the event.

staab:
Cool, I like all of this. Maybe left-align the select buttons though. Otherwise, let me know when it's ready for review and I'll take a closer look.

ticruz38:
It should be fine to review now

Have created 2 new components, in the view folder, PublishesNotices and PublishesConnections, the plan was to have them link to the router but it proved more challenging than expected with the tabs. So these 2 views are just 2 child components for Publishes.

I can move them to another folder if it makes more sense

src/app/views/PublishesConnections.svelte
```diff
@@ -0,0 +1,144 @@
...
+          ).set(url, cxn)
+        } else {
+          newConnectionStatus["Connected"] = (newConnectionStatus["Connected"] || new Map()).set(
+            url,
+            cxn,
+          )
+        }
+      }
+      connectionsStatus = newConnectionStatus
+    }, 800)
```
staab, line 79: This results in a flash of zeroes on first load. Instead, put this in a throttled function and call it on mount and in an interval.

src/app/views/PublishesConnections.svelte
```diff
@@ -0,0 +1,144 @@
...
+      for (const [url, cxn] of ctx.net.pool.data.entries()) {
+        if (pendingStatuses.includes(cxn.auth.status)) {
+          newConnectionStatus["Logging in"] = (newConnectionStatus["Logging in"] || new Map()).set(
+            url,
+            cxn,
+          )
+        } else if (failureStatuses.includes(cxn.auth.status)) {
+          newConnectionStatus["Failed to log in"] = (
+            newConnectionStatus["Failed to log in"] || new Map()
+          ).set(url, cxn)
```
staab, line 54: `pushToMapKey` is what I usually use in situations like this.

src/app/views/PublishesConnections.svelte
```diff
@@ -0,0 +1,144 @@
...
+    AuthStatus.PendingResponse,
+  ]
+
+  const failureStatuses = [AuthStatus.DeniedSignature, AuthStatus.Forbidden]
+
+  $: connections = Array.from(ctx.net.pool.data.entries())
+    .filter(([url, cxn]) =>
+      selectedOptions.length ? selectedOptions.some(s => connectionsStatus[s]?.has(url)) : true,
+    )
+    .map(([url, cxn]) => cxn)
```
staab, line 39: `values` and `filter(cxn => whatever(cxn.url))` would be simpler

src/app/views/PublishesConnections.svelte
```diff
@@ -0,0 +1,144 @@
...
+
+  const options = [
+    "Connected",
+    "Logging in",
+    "Failed to log in",
+    "Failed to connect",
+    "Waiting to reconnect",
+    "Not connected",
+    "Unstable connection",
+  ]
```
staab, line 25: Make this an enum or constants since the strings are used in a bunch of places here. Also, since this is the same logic used in `RelayStatus` we should pull it out into utility functions somewhere.

src/app/views/PublishesConnections.svelte
```diff
@@ -0,0 +1,144 @@
...
+            </span>
+          {/if}
+          <span>
+            Connected {quantify(relay?.stats?.open_count || 0, "time")}
+          </span>
+        </div>
+      </div>
+      <div class="flex w-full items-center justify-end gap-2 text-sm">
+        {#each options as opt}
+          {#if connectionsStatus[opt]?.has(cxn.url)}
```
staab, line 129: Conditionals inside a loop inside a flex container with `gap` will result in odd spacing. Filter `options` instead.

src/app/views/PublishesNotices.svelte
```diff
@@ -0,0 +1,82 @@
...
+  const searchConnections = fuzzy(Array.from(ctx.net.pool.data.keys()))
+
+  function getNotices(connections: Connection[]) {
+    return Object.values($thunks).filter(t =>
+      connections.some(cxn => get(t.status)[cxn.url]?.status),
+    )
+  }
+
+  function messageAndColorFromStatus(status: ThunkStatus) {
+    console.log("message", status)
```
staab, line 24: console log

src/app/views/PublishesNotices.svelte
```diff
@@ -0,0 +1,82 @@
...
+
+  export let selected: string[] = []
+
+  const searchConnections = fuzzy(Array.from(ctx.net.pool.data.keys()))
+
+  function getNotices(connections: Connection[]) {
+    return Object.values($thunks).filter(t =>
+      connections.some(cxn => get(t.status)[cxn.url]?.status),
+    )
+  }
```
staab, line 21: I'd also like to listen for NOTICE messages, which won't come through on thunks. Probably add this via the pool `init` event so we can get handlers on connections when they're set up.

staab (changes requested):

src/app/views/PublishesConnections.svelte
ticruz38: it is not updated inside the loop, the connectionStatus is updated once every 800 ms, there is no flashing on each connection state

staab: On first page load there is, because the interval doesn't run at first:
https://github.com/user-attachments/assets/2b47224e-536d-4af5-a663-510c9bab2546

src/app/views/PublishesNotices.svelte
```diff
@@ -0,0 +1,119 @@
...
+  import {fly} from "svelte/transition"
+  import SearchSelect from "src/partials/SearchSelect.svelte"
+  import AltColor from "src/partials/AltColor.svelte"
+  import {fuzzy} from "src/util/misc"
+  import {router} from "src/app/util"
+  import {subscriptionNotices, type SubscriptionNotice} from "src/engine"
+
+  export let selected: string[] = []
+
+  $: subNotices = selected.flatMap(s => $subscriptionNotices.get(s)?.map(n => ({...n, url: s})))
```
staab, line 15: Maybe derive this in engine/state alongside subscriptionNotices

src/app/views/PublishesNotices.svelte
```diff
@@ -0,0 +1,119 @@
...
+
+  $: pubNotices = getPubNotices(
+    selected.map(url => ctx.net.pool.data.get(url)).filter(Boolean),
+  ) as Thunk[]
+
+  $: notices = ([...pubNotices, ...subNotices] as (Thunk & (SubscriptionNotice & {url: string}))[])
+    .sort((a, b) => {
+      return a.created_at - b.created_at
+    })
+    .filter(Boolean)
```
staab, line 63: `sortBy(x => -x.created_at, ...)` is a more succinct way to do this. Also, the `filter(Boolean)` shouldn't be necessary here, something is wrong upstream if it's needed. `(Thunk & (SubscriptionNotice & {url: string}))` is pretty ugly too, SubscriptionNotice should probably include `url` in its type, and a `|` would be more appropriate instead (although it would require type assertions, so `&` is fine here if it works).

src/engine/model.ts
```diff
@@ -43,3 +43,13 @@ export type AnonymousUserState = {
...
+
+export enum ConnectionType {
+  Connected = "Connected",
+  Logging = "Logging in",
+  LoginFailed = "Failed to log in",
+  ConnectFailed = "Failed to connect",
+  WaitReconnect = "Wainting to reconnect",
+  NotConnected = "Not connected",
+  UnstableConnection = "Unstable connection",
+}
```
staab, line 55: I prefer to decouple constants and display logic by adding an additional `displayConnectionType` function.

src/engine/state.ts
```diff
@@ -1052,6 +1064,17 @@ if (!db) {
     ctx.app.dufflepudUrl = getSetting("dufflepud_url")
   })
 
+  ctx.net.pool.on("init", (connection: Connection) => {
+    // if (!connection.url.includes("snort")) return
+    connection.on(ConnectionEvent.Receive, function (cxn, [verb, ...args]) {
+      if (verb == "EVENT") return
+      subscriptionNotices.update($notices => {
+        pushToMapKey($notices, connection.url, {created_at: now(), notice: [verb, ...args]})
```
staab, line 1072: I'd add `url` to the object just to make the types easier.

staab (commented):
Looking good, a few things:

- https://github.com/coracle-social/coracle/pull/487#discussion_r1866245327 still needs to be resolved
- On the notices page, let's replace the relay selector with a generic "search" box. This should match against relay url, verb (ok/notice), and message.
- The animation on the notices page is janky, use `in` rather than `transition`.
- While I like the granularity on the notices page, I agree with your point that storing all this is going to be pretty heavy, for not very much benefit. Let's slim it down to OK and NOTICE. 
- Let's factor the thunks page out into its own component as well.

ticruz38:
- Lots of refactoring.
- The notice page now shows all notifications; you can filter them afterwards by relay URL, request verb, or connection message.
- When clicking on a connection, it navigates to the notice tab with the search field pre-filled with the clicked relay URL

src/app/views/Publishes.svelte
```diff
@@ -43,28 +51,35 @@
...
+      <span class="text-sm">Succeeded</span>
+    </Tile>
+    <Tile background>
+      <p class="text-lg sm:text-2xl">{recent.length - pending.length - success.length}</p>
+      <span class="text-sm">Failed</span>
+    </Tile>
+  </div>
+  {#each sortBy(t => -t.event.created_at, recent) as thunk (thunk.event.id)}
+    <PublishCard {thunk} />
+  {/each}
```
staab, line 80: Can we move this into a separate component?

src/app/views/PublishesNotices.svelte
```diff
@@ -0,0 +1,89 @@
...
+
+  export let search: string = ""
+
+  $: subNotices = Array.from($subscriptionNotices.values())
+    .flatMap(n => n)
+    .filter(
+      n =>
+        n.url.toLowerCase().includes(search.toLowerCase()) ||
+        n.notice.some(n => String(n).toLowerCase().includes(search.toLowerCase())),
+    )
```
staab, line 21: Use `createSearch` from welshman/app. Also, let's match against the actual text that is displayed to avoid unexpected false negatives. That'll require deriving all items with display information, then feeding that into search.

src/engine/state.ts
```diff
@@ -1058,6 +1067,20 @@ if (!db) {
     ctx.app.dufflepudUrl = getSetting("dufflepud_url")
   })
 
+  ctx.net.pool.on("init", (connection: Connection) => {
+    connection.on(ConnectionEvent.Receive, function (cxn, [verb, ...args]) {
+      if (verb == "EVENT" || verb == "EOSE" || verb == "NEG-MSG") return
```
staab, line 1072: Should be `EVENT|EOSE|CLOSED`

staab: Also, triple equals (humor me)

staab (commented):
Let's actually add `CLOSED` back in (my mistake).

src/engine/state.ts
ticruz38: > Let's actually add CLOSED back in (my mistake)
Those are the events excluded. If I replace NEG-MSG with CLOSED we won't see CLOSED message.

src/app/views/Publishes.svelte
ticruz38: I avoid splitting the views atm, I don't know where to place or how to name those partials component, the list is gigantic. 
Moving to Sveltekit router in a future version would give us more structure and options for organising those components.

src/engine/state.ts
staab: Oh, right, absolutely. Can we make this a white-list instead then? Just list the verbs we do want (since new ones show up on occasion)

ticruz38:
- added a new env variable, LOG_VERBS, that will be visible in the notice console

staab:
What's the use case for putting verbs in .env? I think we can just include them inline. 

One other thing, threshold for the search should probably be higher:

Screenshot 2024-12-09 at 10 41 02 AM

ticruz38:
> What's the use case for putting verbs in .env? I think we can just include them inline.
To make them configurable essentially, I did not know exactly which verbs should be whitelisted

Issue #490, opened 2024-11-25
Prefix entities with `nostr:`


staab:
When creating note `content`

PR #491, opened 2024-11-26
Prefix all entities with nostr:

ticruz38:
It's a small config in nostr-editor extensions, just extend renderText to add `nostr:` in front of nevent, naddr, nprofile.

If we expect those entities to always be prefixed by `nostr:`it might be worth a PR on nostr-editor


ticruz38:
#490

PR #494, opened 2024-11-27
Sending scheduled note via dvm

ticruz38:
This is working, my DVM are picked up by shipyard but the UX is very poor at the moment.

I cannot query back the kind 5905 notes from any of the relay it was published to, I cannot consume the shipyard API without being logged in too. As a result I can't show you how many and which of your note are being scheduled or cancel them.

All I can do is effectively schedule the publishing of your note. This works.


ticruz38:
@staab I tried using the api via a proxy to avoid cors, but shipyard also set domain restricted cookies.

Shipyard really is just an API picking up DVM request. I have 2 options:
- Using shipyard DVM in Coracle, but the UI will be limited.
- Using another schedule DVM, A Coracle one?

ticruz38:
such an API wouldn't be too hard to build, just a server accepting signed event that will be posted on behalf of the signer at the time set in the created_at field of the event propagated and to the relay defined in the tags of the event.

staab:
I don't think we need anything too fancy. And I'd like to avoid relying on a third party service anyway (apart from whatever DVM picks up the request). A UI that allows a user to schedule something in the future, and which makes the DVM request, showing progress and success/failure should be enough.

ticruz38:
- Add the option to schedule the note, in the NoteOptions component, you can pick a date and an Hour.
- Show "Publishing at" instead of "Send" in the submit button in NoteCreate.
- The feedback is similar to a normal send, first the delay, then the statuses, but in the case of a DVM, it will also tell you if the note has been picked up by the DVM in a third toast.

At the moment, only shipyard provide this service, so we could send the user to shipyard.pub (in the toast feedback for example) with a high level of confidence that he will be able to manage its notes there, but maybe on the long run there will be more providers so I left it out for now.

src/app/shared/NoteOptions.svelte
```diff
@@ -29,6 +42,14 @@
       <Field icon="fa-warning" label="Content warnings">
         <Input bind:value={values.warning} placeholder="Why might people want to skip this post?" />
       </Field>
+      {#if publishAt}
+        <Field icon="fa-hourglass-half" label="Schedule post">
+          <Input
+            type="datetime-local"
+            min={new Date().toISOString().slice(0, 16)}
+            bind:value={values.publish_at} />
+        </Field>
```
staab, line 51: We have a DateTime input that should simplify some of the parsing/formatting logic

src/app/views/NoteCreate.svelte
```diff
@@ -77,20 +103,55 @@
       tags.push(tagPubkey(quote.pubkey))
     }
 
-    const template = createEvent(1, {content, tags})
+    const template = createEvent(1, {
+      content,
+      tags,
+      created_at: options?.publish_at,
```
staab, line 109: No need for `?`, options should always be defined. Same thing below.

src/app/views/NoteCreate.svelte
```diff
@@ -30,10 +48,18 @@
...
+  let options = {warning: "", anonymous: false, publish_at: null}
+
+  const DVM_HANDLER_FILTER = {
+    kinds: [HANDLER_INFORMATION],
+    "#k": [DVM_REQUEST_PUBLISH_SCHEDULE.toString()],
+  }
+
+  const handlers = deriveEvents(repository, {
+    filters: [DVM_HANDLER_FILTER],
+  })
```
staab, line 62: Instead of loading handlers, let's just hardcode shipyard's pubkey (since we don't allow users to select which dvm to use anyway)

src/app/views/NoteCreate.svelte
```diff
@@ -113,11 +174,19 @@
     thunk.status.subscribe(status => {
       if (
         Object.values(status).length === thunk.request.relays.length &&
-        Object.values(status).every(s => s.status === PublishStatus.Pending)
+        Object.values(status).every((s: any) => s.status === PublishStatus.Pending)
```
staab, line 177: This is the wrong way to handle this, instead use `thunk: Thunk` where it's declared

src/app/views/NoteCreate.svelte
```diff
@@ -113,11 +174,19 @@
...
       ) {
         showPublishInfo(thunk)
       }
     })
+    if (emitter) {
+      emitter.on(DVMEvent.Progress, (url: string, event: TrustedEvent) => {
+        $signer.nip04.decrypt(dvmPubkey, event.content).then(data => {
+          data = JSON.parse(data)[0]
+          showInfo(data[2] || "You note is " + data[1])
```
staab, line 186: Try catch this, and add a `!` at the end of the message (also fix the typo, you -> your)

src/app/views/NoteCreate.svelte
```diff
@@ -138,16 +207,19 @@
...
+    const options = getEditorOptions({
+      content: drafts.get("notecreate") || "",
+      submit: onSubmit,
+      element,
+      submitOnEnter: false,
+      submitOnModEnter: true,
+      autofocus: true,
+    })
+
+    editor = new Editor(options)
```
staab, line 222: No need to shadow `options`, just pass it straight in

src/app/views/NoteCreate.svelte
```diff
@@ -169,6 +241,10 @@
   })
 
   $: loading = editor?.storage.fileUpload.loading
+
+  $: dvmPubkey = $handlers[0].pubkey
```
staab, line 245: This is crashing if handlers isn't defined. I mentioned this above, but let's just hardcode the dvm pubkey.

src/app/views/NoteCreate.svelte
```diff
@@ -169,6 +241,10 @@
...
+
+  $: console.log($handlers)
```
staab, line 247: debug code

src/app/views/NoteCreate.svelte
```diff
@@ -217,6 +293,14 @@
...
           {#if signaturePending}
             <i class="fa fa-circle-notch fa-spin" />
+          {:else if options?.publish_at && options?.publish_at > now()}
+            {#if new Date(options.publish_at * 1000).toDateString() === new Date().toDateString()}
+              Scheduled for {new Date(options.publish_at * 1000).toLocaleTimeString()}
+            {:else}
+              Scheduled for {new Date(options.publish_at * 1000).toLocaleDateString()} at {new Date(
+                options.publish_at * 1000,
+              ).toLocaleTimeString()}
+            {/if}
```
staab, line 303: Instead of lots of all caps and numbers, let's just change this to "Schedule"

src/app/views/NoteCreate.svelte
```diff
@@ -227,12 +311,18 @@
           <i class="fa fa-upload" />
         </button>
       </div>
+      <button
+        type="button"
+        class="flex cursor-pointer items-center justify-end gap-4 text-sm"
+        on:click={() => (showOptions = true)}>
+        <span><i class="fa fa-warning" /> {options.warning || 0}</span>
+      </button>
```
staab, line 319: This is redundant with the cog icon above, I removed this recently.

src/engine/commands.ts
```diff
@@ -101,6 +104,30 @@ export const nip98Fetch = async (url, method, body = null) => {
...
+      emitter.emit(DVMEvent.Result, url, event)
+
+      if (autoClose) {
+        sub.close()
+      }
+    }
+  })
+  return {request, emitter, sub, thunk}
+}
+
```
staab, line 130: Why not use `@welshman/dvm`?

src/engine/state.ts
```diff
@@ -221,6 +222,8 @@ const unwrapper = new Worker<TrustedEvent>({chunkSize: 10})
 unwrapper.addGlobalHandler(async (event: TrustedEvent) => {
   if (event.kind === WRAP) {
     await ensureUnwrapped(event)
+  } else if (event.kind === DVM_REQUEST_PUBLISH_SCHEDULE) {
+    console.log("DVM_REQUEST_PUBLISH_SCHEDULE", event)
```
staab, line 226: debug code

staab (changes requested):

src/engine/commands.ts
ticruz38: Just to get a Thunk with the delay, the toaster also understand thunk better to display the request status

src/app/views/NoteCreate.svelte
ticruz38: I suggest to define the DVM pubkey in .env, some people seems to be mad at Shipyard

Hardcoding for now

src/app/views/NoteCreate.svelte
ticruz38: maybe add a small text below the button for when it is actually scheduled, just the date in local date format, ie 12-23-2025 at 20:00

src/app/views/NoteCreate.svelte
ticruz38: I don't understand this comment, here I am checking that all thunk request status are pending before sending the thunk status toaster.
Would you rather send it with different requirements, ie one request status being set at least?

staab: Sorry, I mean the type hint. The line where thunk is assigned should be the line to be fixed, the whole file thinks thunk is `any`

src/app/views/NoteCreate.svelte
staab: Let's leave it off for now, until people ask for it. They just set the date, so the confirmation shouldn't be necessary

src/engine/commands.ts
staab: Gotcha, I guess that's ok

PR #495, opened 2024-11-28
Refine Modals

ticruz38:
Modal Drawer is now the default except for HUD, profiles and feed Edit

The side drawer has a max-width-md applied, opens from the left, the sidebar is still visible and accessible
Screenshot 2024-11-28 at 17 45 11

When another drawer opens it fully covers the first one
Screenshot 2024-11-28 at 17 45 21

When opening a profile page, we still have the former modal, but the side menu is still visible, I thought it was ok, but I can put it in the background if needed
Screenshot 2024-11-28 at 17 46 03

On mobile the modal still open from the bottom, I can open it from the side if needed, it is now dismissible with a swap.

https://github.com/user-attachments/assets/579e199a-f005-4927-8c40-80a1d3eef66a

### DIFF
--- a/src/app/MenuDesktop.svelte
+++ b/src/app/MenuDesktop.svelte
@@ -71,7 +71,7 @@
   $: userDisplay = deriveProfileDisplay($pubkey)
 </script>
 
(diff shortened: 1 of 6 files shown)


staab:
Rebased and merged

PR #496, opened 2024-11-29
Fix ascii tags

ticruz38:
Screenshot 2024-11-29 at 14 41 06

The proof right here 👆


ticruz38:
Tag extension now uses suggestions

https://github.com/user-attachments/assets/b89dc213-ea18-4e3c-8fd3-9bd65fa3fe00

https://github.com/user-attachments/assets/5d446ef9-4070-4408-9be0-9a397030ca39

Issue #455, opened 2024-10-29
consider: do not show "open other nostr app" icon if viewing/opening kind1s


alltheseas:
_context_ 

Without knowing if Coracle users use the "open note in another app" link (seems unlikely, as Coracle does fine with kind1s), consider not displaying this option for notes. Instead, display for less commonly supported kinds/kinds not supported by Coracle.

https://njump.me/note15nafwgnws46p0kr373suf7vyg8j3un38p3gy3ukws6prj6ctp3as8z4vtw

alltheseas:
![kind1coracle](https://github.com/user-attachments/assets/d419356d-f513-4d5b-bf5a-f98bc840c8af)

staab:
The rendering architecture needs to change, since we switch on kind only within `NoteContent`. With e.g. feeds, it will make more sense to swap out rendering for the entirety of the note instead. Kind handlers should only show up on non-kind-1 notes, and probably be integrated more tightly with the content. Also use a different icon.

alltheseas:
Different icon via Niel "open with"

![image](https://github.com/user-attachments/assets/a0b1b3b1-2c62-44a1-ae8a-bc7effaac8ea)

https://njump.me/nevent1qqst6r4q96mqysfkujexzsfm6hy3w4ht64a45mmr5wxj9wkxsjac2jgpz3mhxue69uhkummnw3ezummcw3ezuer9wcq32amnwvaz7tmwdaehgun9d35hgetn9ehhyecpzfmhxue69uhhwmm59e6hg7r09ehkuegpr9mhxue69uhhqatjv9mxjerp9ehx7um5wghxcctwvsqf6vy0

PR #497, opened 2024-11-29
Fix `open with` for note kind #1

ticruz38:
#455 

> The rendering architecture needs to change, since we switch on kind only within NoteContent. With e.g. feeds, it will make more sense to swap out rendering for the entirety of the note instead. Kind handlers should only show up on non-kind-1 notes, and probably be integrated more tightly with the content. Also use a different icon.

I don't have the context or state of Coracle at the time of this comment but from what I can see in the code, I have some questions:
- You do still switch NoteContent based on kind.
- By **entirety of the note** do you mean having NoteActions embedded in NoteContent? I like both being separated still.
- Kind Handlers are tightly integrated with the note content, as it seems each handlers has a kind "attached" to it. You are already switching handlers based on the Note kind. Or you want to use another means, aside from the note kind, to filter out the Note handler?
- The icon choice is quite limited on Font-Awesome, I came up with the fa-list icon, which seems more appropriate, for a wider range of icon I personally use iconify, but that's a big departure from font-awesome, even though iconify support font-awesome too. It does add some dependencies that you need to be confortable with.


staab:
The iconify version looks better, but we shouldn't add a dependency. See if you can get something like that into `Icon.svelte` in svg format (although with an arrow would be preferable). I often use https://www.flaticon.com/ for stuff like this.

For an illustration of what I mean, certain note kinds should ideally be able to take over the layout of the profile picture etc. to differentiate them from normal notes. Here's how a feed event currently looks:

Screenshot 2024-12-02 at 9 40 55 AM

Which obviously is not ideal. Here's an example of how it might look:

Screenshot 2024-12-02 at 9 46 25 AM

This is going to be very hairy to accomplish, since it'll require allowing NoteContentX components to call NoteActions internally, or maybe use crazy slots. Then, of course, there are quotes, which are already convoluted to avoid circular dependencies. The simplest solution might be to add logic to Note.svelte, but that's going to be ugly too. I'd really like the solution to be very clean, since there is so much mission-critical stuff going on in these files.

ticruz38:
What could be nice trying:

- Use svelte-infinite-scroll. This component allows us to determine what is visible on screen, which might give us an opportunity to optimise performance by only requesting/computing information for the elements on-screen or nearby. I'm thinking mostly of NoteActions.svelte, this one is probably the most expensive.

src/app/shared/Feed.svelte
```diff
@@ -240,10 +240,10 @@
   </FeedControls>
 {/if}
 
-<FlexColumn xl bind:element>
+<FlexColumn bind:element>
   {#each events as note, i (note.id)}
     <div in:fly={{y: 20}}>
-      <Note {filters} {reposts} {depth} {anchor} {note} />
+      <Thread {filters} {reposts} {depth} {anchor} {note} />
```
staab, line 246: Thread sounds like ThreadDetail, which is a different kind of thing. Let's name this `FeedItem`

staab (commented):
Interesting, looks fine so far, but I'm curious to see what it would look like applied to specific note kinds (e.g. feeds)

staab:
I've gone ahead and bumped this update to 0.6.0 since it's more substantial than I was expecting.

ticruz38:
This PR primarily refactor the Note.svelte component now renamed FeedItem.svelte

Note.svelte now represent just a single note in the feed and is divided into 3 parts: header, content and actions (footer). 
It is now easier to fine tune the display depending on the note kind.

The "Open With" action now uses Niels' icon.

staab:
Looks pretty good, can you fix conflicts? Also, can we get the icon to be a little thicker? The lines are skinnier than font-awesome's, so it looks sort of disabled.

ticruz38:
rebased dev and increase open-with stroke width

src/app/shared/NoteReply.svelte
```diff
@@ -1,61 +1,54 @@
...
 
-  const setOptions = values => {
-    options = {...options, ...values}
-    showOptions = false
+  const openOptions = () => {
+    showOptions = true
+  }
+
+  const setOpts = e => {
+    opts = {...opts, ...e.detail}
```
staab, line 51: Looks like you've undone some of my changes, options are no longer working correctly on replies

src/app/shared/NoteReply.svelte
```diff
@@ -163,14 +143,10 @@
...
+    <!-- {#if showBorder}
       <AltColor background class="absolute -top-4 z-none h-5 w-1" />
-    {/if}
-    <div
-      transition:slide|local
-      class="note-reply relative my-2 gap-1"
-      bind:this={container}
-      on:click|stopPropagation>
+    {/if} -->
+    <div transition:slide|local class="note-reply relative my-2 gap-1" on:click|stopPropagation>
```
staab, line 149: This looks unresolved

src/app/shared/NoteActions.svelte
```diff
@@ -150,31 +171,50 @@
...
+  $: handlers =
+    event.kind != 1 &&
+    $kindHandlers.filter(
+      h =>
+        h.name.toLowerCase() !== "coracle" &&
+        h.event.tags.some(
+          t =>
+            ["web", os].includes(t[0]) && (t.length == 2 || ["note", "nevent"].includes(last(t))),
+        ),
+    )
```
staab, line 217: There are important changes in this conflict, the empty mark is needed for certain handlers.

src/app/shared/NoteReply.svelte
```diff
@@ -53,8 +47,8 @@
     showOptions = false
   }
 
-  const setOptions = values => {
-    options = {...options, ...values}
+  const setOpts = values => {
+    opts = {...opts, ...values}
```
staab, line 51: Can you keep my rename please

src/app/shared/NoteReply.svelte
```diff
@@ -149,11 +131,10 @@
     editorLoading = editor.storage.fileUpload.loading
   }
 
-  $: {
-    if (editorElement) {
-      createEditor()
-    }
-  }
+  // eslint-disable-next-line
+  $: editorElement && createEditor()
```
staab, line 135: Keep my version please, I want to avoid eslint directives

staab (commented):
Just a couple small things, when resolving conflicts try to keep the changes made upstream rather than revert them

src/app/shared/NoteActions.svelte
ticruz38: blaming my pink eye for it

PR #503, opened 2024-12-10
Chat send is slow

ticruz38:
There was an unnecessary roundtrip to the signer when sending a message. We were trying to unwrap a message we had just wrapped. This was fixed by publishing the rumor to the local repository before publishing the wrapped version to the relays.

The editor content is now cleared as soon as the message is sent. A loader replaces the send icon while the message is being wrapped and signed. Once the message is ready to be published, it appears in the channel, where it can still be canceled during the delay set by the user in the app settings.

https://github.com/user-attachments/assets/85ce862f-b65f-45d2-8dc0-6850593da574

### DIFF
--- a/src/app/views/ChannelsDetail.svelte
+++ b/src/app/views/ChannelsDetail.svelte
@@ -10,7 +10,7 @@
   import PersonCircles from "src/app/shared/PersonCircles.svelte"
   import PersonAbout from "src/app/shared/PersonAbout.svelte"
   import {router} from "src/app/util/router"
-  import {sendMessage, markChannelRead, getChannelIdFromEvent, listenForMessages} from "src/engine"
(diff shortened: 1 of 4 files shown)


staab:
Looking better:

- Go ahead and use the same loading icon used elsewhere (fa-circle-notch)
- See if you can use publishThunk and get the message to show up immediately
- On cancel, does the note content get put back into the editor?

ticruz38:
> See if you can use publishThunk and get the message to show up immediately

It is using publishthunk, but the wrapped message is sent and saved to the repository, publishthunk do work well, it's just that I'm unwrapping an incoming message for which I know the rumor already.

> On cancel, does the note content get put back into the editor?

Yes 

https://github.com/user-attachments/assets/aa550a6a-49d2-4dd3-93cb-1cb745f47b57

Issue #500, opened 2024-12-03
Fix notifications


staab:
Currently notifications are implemented using [this draft](https://github.com/nostr-protocol/nips/pull/1405)

They unfortunately tend to blink in and out as messages are being retrieved. We could try to fix the NIP, but there are a number of problems with it. For now, let's remove read status events, and store read status locally. See flotilla for an example of how this is done there.

PR #504, opened 2024-12-11
Use local read receipts for notifications

ticruz38:
#500

Use a local store for read receipts, inspired by flotilla.


staab (commented):
Looks pretty good, although the notifications utilities should go in `engine/notifications` rather than domain (domain is purely for stateless utility functions related to nostr data types).

src/engine/state.ts
```diff
@@ -373,120 +369,19 @@ export const isEventMuted = withGetter(
...
-  getSeenAt,
-  $getSeenAt => (key: string, event: TrustedEvent) => $getSeenAt(key, event) > 0,
-)
+export const deriveChecked = (key: string) => derived(checked, prop(key))
 
-// Notifications
+export const getSeenAt = derived([checked], ([$checked]) => (key: string, event: TrustedEvent) => {
+  const match = $checked[key]
+  const fallback = $checked["*"]
+  const ts = Math.max(match || 0, fallback || 0)
```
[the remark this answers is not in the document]
ticruz38: welshman has so many secrets

src/app/state.ts
```diff
@@ -29,6 +29,7 @@ import {
   userFeedFavorites,
   getSetting,
 } from "src/engine"
+// import {setChecked} from "src/domain/notifications"
```
staab, line 32: This looks incomplete

src/app/views/NotificationSectionMain.svelte
```diff
@@ -34,31 +34,24 @@
     })
   })
 
-  let loading = false
-
   onMount(() => {
     const tracked = new Set()
+    setChecked(["replies", "mentions"])
 
     const unsub = unreadMainNotifications.subscribe(async events => {
```
staab, line 41: This is all dead code now

src/app/views/NotificationSectionReactions.svelte
```diff
@@ -32,30 +32,24 @@
     })
   })
 
-  let loading = false
-
   onMount(() => {
     const tracked = new Set()
+    setChecked(["reactions", "zaps"])
 
     const unsub = unreadReactionNotifications.subscribe(async events => {
```
staab, line 39: Also dead

src/app/views/ChannelsList.svelte
```diff
@@ -39,6 +39,8 @@
     }
   })
 
+  const markAllChannelsRead = () => setChecked($channels.map(c => c.id))
```
staab, line 42: Let's do something here like `channels/*`, since old events frequently come in after marking all as read, resetting the notification bell. This is a flaw of the current approach.

staab (commented):
- Various `SEEN_` events are still being fetched and decrypted, we can remove all of that.
- The subscriptions on the notifications pages aren't doing anything any more.
- Record "mark all as seen" explicitly to avoid getting re-notified about old stuff. You'll need to check that when getting notifications for channels, which means you need to know which notification keys correspond to channels. So a different naming scheme than channel id might be needed.

ticruz38:
- Using a "channels/" prefix as a seen key for messages
- seen all message is now dispatched as channels/* instead of a key for each channel.
- Removed subscriptions in notifications onMount event
- Set `all read` when logged in and signed up

staab:
This can still be simplified a decent amount I think:

- Since this is route based, `setChecked` can take a single key (like in flotilla). There's no need to differentiate between reactions/zaps for example, that was added for theoretical interoperability.
- `*` should apply to all notifications, including channels, `channels/*` is a subset of that

ticruz38:
applied your suggestions, path based seen check

Notifications are now under notes/* or reactions/*, everything related to messages is under channels/*, "*" is used for login and signup and is a fallback for any subpath no checked yet.

Issue #372, opened 2024-06-20
Get rid of ramda


staab:
Pretty much everything ramda provides is also provided by welshman/lib. Be careful though, because the function signatures are sometimes different! If an equivalent function doesn't exist, just write a vanilla js version.

PR #508, opened 2024-12-13
replace ramda with welshman/lib

ticruz38:
#372 

Replaced most of ramda utilities with either welshman/lib or standard JS functions.

takeWhile might deserve its own implementation in welshman/lib.


src/app/App.svelte
```diff
@@ -353,7 +352,7 @@
     } else if (document.body.style.position === "fixed") {
       document.body.setAttribute("style", "")
 
-      if (!isNil(scrollY)) {
+      if (scrollY) {
```
staab, line 355: This is important in the case of scrollY being 0; welshman/lib has an isNil function

src/app/Routes.svelte
```diff
@@ -74,7 +73,7 @@
   </div>
 {/key}
 
-{#each reverse($modals).filter(m => !m.virtual) as m, i (router.getKey(m) + i)}
+{#each $modals.reverse().filter(m => !m.virtual) as m, i (router.getKey(m) + i)}
```
staab, line 76: This mutates the variable, which is definitely not what we want.

src/app/shared/Channel.svelte
```diff
@@ -86,12 +85,12 @@
   const scrollToBottom = () => element.scrollIntoView({behavior: "smooth", block: "end"})
 
   const stickToBottom = async () => {
-    const lastMessage = pluck("created_at", groupedMessages).reduce(max, 0)
+    const lastMessage = max(groupedMessages.map(prop("created_at")) as number[])
```
staab, line 88: Prefer lambdas over type assertions. Prop is probably a bad idea.

src/app/views/Zap.svelte
```diff
@@ -17,60 +15,53 @@
   export let splits
   export let id = null
   export let anonymous = false
-  export let amount = getSetting("default_zap")
+  export let amount = getSetting("default_zap") as string
```
staab, line 18: Assertion is incorrect, it should be a number right?

src/engine/state.ts
```diff
@@ -562,7 +566,7 @@ export const withIndexers = (relays: string[]) => withRelays(relays, env.INDEXER
 export const lists = deriveEventsMapped<PublishedUserList>(repository, {
   filters: [{kinds: EDITABLE_LIST_KINDS}],
   eventToItem: (event: TrustedEvent) => (event.tags.length > 1 ? readUserList(event) : null),
-  itemToEvent: prop("event"),
+  itemToEvent: prop("event") as any,
```
staab, line 569: Use a lambda here instead of `any`

src/engine/state.ts
```diff
@@ -726,7 +730,7 @@ export const collectionSearch = derived(
 // Network
 
 export const getExecutor = (urls: string[]) => {
-  const [localUrls, remoteUrls] = partition(equals(LOCAL_RELAY_URL), urls)
+  const [localUrls, remoteUrls] = partition(url => equals(LOCAL_RELAY_URL, url), urls)
```
staab, line 733: `===` is fine here, equals doesn't add anything for strings

src/util/router.ts
```diff
@@ -208,9 +208,9 @@ class RouterExtension {
     return this.clone({queryParams: data})
   }
 
-  cx = context => this.clone(updateIn("context", mergeLeft(context))(this.params))
+  cx = context => this.clone(updateIn("context", (c: object) => ({...c, ...context}))(this.params))
 
-  cg = config => this.clone(updateIn("config", mergeLeft(config))(this.params))
+  cg = config => this.clone(updateIn("config", (c: object) => ({...c, ...config}))(this.params))
```
staab, line 213: mergeLeft is available in welshman

staab (commented):
- I've released a new version of welshman with `pluck` in it, go ahead and use that instead of map/prop. Also `spec` and `complement` can be useful to replace `whereEq`, but that's optional
- Prop is now parameterizable, so you can avoid type assertions
- I've added a linting rule for triple equals, be sure to rebase and lint

src/app/views/Zap.svelte
ticruz38: yes

src/app/App.svelte
```diff
@@ -435,7 +432,7 @@
         ])
 
         if (lastActivity && lastActivity < ago(30)) {
-          ctx.net.pool.get(connection.url).socket.close()
+          ctx.net.pool.remove(connection.url)
```
staab, line 435: Nope, don't revert this

staab (commented):
One final thing and it's good to go

staab:
Merged manually

Issue #492, opened 2024-11-26
Ask before loading/decrypting messages


staab:
The first time a user visits the messages tab, ask them if they want to turn on message decryption. Don't load 4/1059 until they do. See Flotilla for an example of how this should work.

PR #511, opened 2024-12-17
Enable decryption

ticruz38:
#492

Do not immediately unwrap messages until the user explicitly say it by visiting the message page at least once.

### DIFF
--- a/src/engine/state.ts
+++ b/src/engine/state.ts
@@ -150,7 +150,7 @@ export const hasNip44 = derived(signer, $signer => Boolean($signer?.nip44))
 
 export const anonymous = withGetter(writable<AnonymousUserState>({follows: [], relays: []}))
 
-export const canDecrypt = synced("canDecrypt", false)
+export const canDecrypt = withGetter(synced("canDecrypt", false))
 
 // Plaintext
 
@@ -238,7 +238,7 @@ repository.on("update", ({added}: {added: TrustedEvent[]}) => {
     }
 
     if (event.kind === WRAP) {
-      if (get(canDecrypt)) {
+      if (canDecrypt.get()) {
         unwrapper.push(event)
       }
     }


src/app/views/ChannelsList.svelte
```diff
@@ -77,3 +79,7 @@
     <Content size="lg" class="text-center">No messages found.</Content>
   {/each}
 </FlexColumn>
+
+{#if !$canDecrypt}
+  <ChatEnable />
+{/if}
```
staab, line 85: For some reason, the modal animation isn't kicking in, maybe because of the history bug mentioned below.

ticruz38: For the modal animation to properly animate you need to have it bound to the router navigation. This is how I'm doing it now

src/app/views/ChatEnable.svelte
```diff
@@ -0,0 +1,57 @@
...
+    canDecrypt.set(true)
+  }
+  const submit = async () => {
+    loading = true
+
+    try {
+      await enableChat()
+    } finally {
+      loading = false
+      router.at("channels").push()
```
staab, line 24: This adds a history item, when we probably want to remove one instead. I think `router.pop()` will be better here.

src/app/views/ChatEnable.svelte
```diff
@@ -0,0 +1,57 @@
...
+      ensureUnwrapped(event)
+    }
+
+    canDecrypt.set(true)
+  }
+  const submit = async () => {
+    loading = true
+
+    try {
+      await enableChat()
```
staab, line 21: enableChat doesn't need to be async, since you're not awaiting ensureUnwrapped (which is fine)

src/engine/state.ts
```diff
@@ -236,7 +238,9 @@ repository.on("update", ({added}: {added: TrustedEvent[]}) => {
     }
 
     if (event.kind === WRAP) {
-      unwrapper.push(event)
+      if (get(canDecrypt)) {
```
staab, line 241: Use `withGetter` when defining the store and `.get()` since this is a fairly hot code path.

staab (changes requested):

staab:
Merged via e0e29cf8

Issue #510, opened 2024-12-16
Code and code blocks not rendered


staab:
When composing a note using nostr-editor that includes code fencing, the backticks don't end up as part of the published note content.

PR #513, opened 2024-12-19
Trying to modify Code extension from mark to node

ticruz38:
#510

CodeBlock is now parsed correctly when rendering to text.

Code on the other hand is almost impossible to correctly transform from Mark (that does not parse to text via renderText()) to a Node (that does parse to text via renderText())

It's frustrating to not be able to do such minor changes easily, leaving the code here if you want to have a look, I'm close but there is a tons of edge case still

### DIFF
--- a/src/app/editor/Code.ts
+++ b/src/app/editor/Code.ts
@@ -1,0 +1,87 @@
+import {inputRegex, pasteRegex, type CodeOptions} from "@tiptap/extension-code"
+import {
+  Mark,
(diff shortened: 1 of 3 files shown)


staab:
This is pretty much what I would expect, marks seem borderline useless to me. I am working on moving the editor into welshman, so I may not merge this, but I will probably copy the changes over.

staab:
I've incorporated this into the editor refactor I'm working on.

Issue #514, opened 2024-12-29
Reply status on small screen


staab:
![Screenshot_20241229-071903](https://github.com/user-attachments/assets/6ba9cf9d-e98b-4327-99ad-54c4311ca6ab)

PR #518, opened 2025-01-08
adjust pending message depending on screen size

ticruz38:
Strip down the reply-status message on mobile:
- Sending Reply in 3 seconds => Sending in 3s
- Published to 1/3 (1 failed, 1 timed out) => Published to 1/3

#514

### DIFF
--- a/src/app/shared/NotePending.svelte
+++ b/src/app/shared/NotePending.svelte
@@ -23,7 +23,8 @@
   import {userSettings} from "src/engine"
   import Anchor from "src/partials/Anchor.svelte"
   import {timestamp1} from "src/util/misc"
-  import {openReplies} from "../state"
+  import {isMobile} from "src/util/html"
+  import {openReplies} from "src/app/state"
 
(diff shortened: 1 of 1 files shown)


src/app/shared/NotePending.svelte
```diff
@@ -56,8 +58,11 @@
 
   $: isPending = pendings > 0
   $: isCompleted = total === success + failed + timeout
+  $: isMobile = innerWidth < 720
```
staab, line 61: util/html has an `isMobile` variable that should work (no real need to be dynamically responsive)

staab (changes requested):

staab:
Merged via fd38c38b

@ticruz38 take a look at the diff, the layout still wasn't great on small screen sizes, so I re-worked it. Main thing to notice is that I used responsive classes to hide/show wording, which are preferable to the isMobile variable when available.

ticruz38:
yes much cleaner via classname

Issue #517, opened 2025-01-06
Long urls break hud


staab:
![Screenshot 2025-01-06 at 12 31 48 PM](https://github.com/user-attachments/assets/ecfcaeb2-4a4d-434a-b020-7fc7efc35ab8)

PR #519, opened 2025-01-08
accomodate long url in hud console

ticruz38:
Very long url will go to the next line in the console, if the url is larger than the container it will result in an ellipsis

Screenshot 2025-01-08 at 17 17 35

#517


dtonon:
@ticruz38 sorry for the delay, it's fine!

Issue #516, opened 2025-01-06
Support kind 20


staab:
add to noteKinds, render

PR #520, opened 2025-01-09
Support kind 20 (Picture First)

ticruz38:
Kind 20 being an instagram like kind of post. Here are the assessment I made for this PR:
- Maintain the original image height in the feed, do not shrink it.
- For posts with multiple images, display them in a carousel format instead of stacking them vertically.
- show image controls on hover
- Full-screen mode is not available.

 #516 

Screenshot 2025-01-09 at 14 44 56

https://github.com/user-attachments/assets/780ecf55-e89c-4ec8-9e77-5f66646fbf0a


staab (changes requested):
- Carousel is spelled with one "r"
- Make sure to hide media if the user has disabled show_media

ticruz38:
This is what a kind 20 post looks like when media is hidden.

Since the picture is central to these posts, it should be clear to the user why no picture is displayed, and a call to action should be suggested., as oppose to kind 1 event who can just display the link.

Screenshot 2025-01-14 at 12 22 25

ticruz38:
PICTURE_FIRST kind 20 needs to be added to welshman for this PR to compile

staab:
Merged via 993ddb20

Issue #522, opened 2025-01-13
Profile description doesn't immediately update


staab:
Edit profile, change description, and it doesn't update on the page. Check other fields too.

PR #535, opened 2025-01-15
key PersonAbout with profile.about

ticruz38:
In PersonAbout, we pass a custom built note to NoteContentKind1, this note is simply muted when changing the about field on profile. So NoteContentKind1 do not re-render.

I tried a bunch of stuff to pass in a brand new note object to NoteContentKind1, nothing worked. I then just wrapped NoteContentKind1 in a #key that is the about field. Not very happy with it.

You probably have a better idea

### DIFF
--- a/src/app/shared/PersonAbout.svelte
+++ b/src/app/shared/PersonAbout.svelte
@@ -8,10 +8,12 @@
   const profile = deriveProfile(pubkey)
 </script>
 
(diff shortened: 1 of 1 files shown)


ticruz38:
#522

staab:
Fixed via 49a5f67d which is cleaner. Hopefully it won't have performance problems.

Issue #523, opened 2025-01-13
Hide reply status on subsequent note views


staab:
It's a fairly common thing for users to create a reply, then reply to their own reply. Currently this is impossible without a page reload. Only show note publish status on replies at first, on subsequent views of the note it should look like a normal note. The pattern to test is:

- Send a reply
- User clicks on the reply, which opens it in a modal
- The reply should have normal actions, not reply status

PR #539, opened 2025-01-16
Hide reply status when the note is being highlighted

ticruz38:
#523

I am using the topLevel property on FeedItem and pass it via context all the way down to NotePending.

This allows me to hide the note status when the note is first in a displayed feed. Which happens naturally when you click on a reply or note.


staab:
Merged via 32aa95c0e82f4aa785186a8427fb6af0d2dff928

Issue #116, opened 2023-08-02
Pinned posts


staab:
That show up at the top of a person detail page. Use kind 10001

PR #540, opened 2025-01-17
Pinned notes

ticruz38:
#116 

Pinned post are now visible on the profile page and can be pinned from the note actions.

Screenshot 2025-01-17 at 12 56 24

It looks like this 👆

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@tsconfig/svelte": "^5.0.4",
     "@types/throttle-debounce": "^5.0.2",
     "@vite-pwa/assets-generator": "^0.2.6",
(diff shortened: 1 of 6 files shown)


ticruz38:
that PR consider that @welshman/app 0.0.39 exist, welshman needs to be bump from 38 to 39 @staab 

You can still test the PR if welshman is linked to coracle on master

staab:
Merged via 45c371a6

staab:
Main thing to note is that PINS weren't getting loaded anywhere, so this would only ever show up for the currently logged in user, and only for notes pinned in the same session.

Issue #167, opened 2023-10-03
Make images nicer


staab:
- When clicking on an image, open it full-screen in-app
- When multiple images are inserted serially, tile them like twitter does

manimejia:
Gon try to tackle this as "low hanging fruit" dip my toes in the water of coracle.

manimejia:
I assume MediaSet.svelte is the template to use. I see it implemented once in NoteContentType1.svelte, supposedly for rendering "extraLinks" that don't fit in the note summary? But I don't see this actually rendering content.  

https://github.com/coracle-social/coracle/blob/04f8d898554ba5511c55fabc196a9c7d12de8e51/src/app/shared/NoteContentKind1.svelte#L77

Is this a bug? maybe MediaSet.svelte is just a WIP for me to work on? or maybe I'm not properly reading the intentions of this code, and there is another way for me to see this template "in action"?

staab:
This is all a little disorganized. MediaSet is only used in one place, and you're right, it might be dead code at this point. Feel free to move things around and re-factor. The main problem is that more kinds of media (audio, link previews, video, quotes) were added over time and conflicted with things like note truncation, show media settings, etc.

We should avoid re-organizing people's media too much, so it would probably best to identify media in sequence and group them.

manimejia:
Hmm... looks like the "isMedia" property of URLs "discovered" in note content does not distinguish between web pages (excluding domain root?) and actual media content...

``` typescript
 export const urlIsMedia = (url: string) => 
!url.match(/\.(apk|docx|xlsx|csv|dmg)/) && last(url.split("://"))?.includes("/") 
```

https://github.com/votegold/coracle/blob/13a1f9d57726a6f580617f184475bb362d7a46b8/src/util/notes.ts#L18-L19

That complicates the matter in our (current and future) efforts to present links to images and other media in a manner visually distinct from HTML web page links. ONLY because I'm NOT smart enough to know the RIGHT regex to use for such a purposes. However I AM smart enough to see that the above code is insufficient for determining the content of outgoing links for our needs. 

So... is there a preferred node module or other library we can use for this? (asking for a friend)

staab:
[...]

PR #546, opened 2025-01-22
Improve images

ticruz38:
- Fullscreen support
- Display images in tiles (like twitter does)
- Works well on mobile

Not ready to review, have a few issues to iron out, but you can already have a good overall idea.

#167 

Screenshot 2025-01-22 at 17 22 01

https://github.com/user-attachments/assets/a190fd5a-698a-4280-92f0-9ace3de4fe0d


ticruz38:
Have modified the note parsing, to aggregate images in a sub-array, this could also be handled in Welshman if it was to be normalised.

ticruz38:
people/nprofile1qyv8wumn8ghj7cmjv4shgu3wdehhxarj9emkjmn99uq3uamnwvaz7tmwdaehgu3wd4shs6tdv93kjarpv3jkctn0wfnj7qgmwaehxw309aex2mrp0yhxvctjwd3kzurfv9hzucm0d5hsqg8hw67vzgn3heulcudk2heuhl4n4zsknue7uye4lnqvy2pfjqw6fgt80mys

Malos is a photographer on Nostr, good place to test out this PR

src/app/MenuDesktop.svelte
```diff
@@ -75,7 +75,7 @@
   $: userDisplay = deriveProfileDisplay($pubkey)
 </script>
 
-<div class="fixed bottom-0 left-0 top-0 z-modal w-72 bg-tinted-700 transition-colors">
+<div class="fixed bottom-0 left-0 top-0 z-nav w-72 bg-tinted-700 transition-colors">
```
staab, line 78: What's the reason for changing this? This breaks the orange overlapping line.

staab (commented):
- Escape closes both the image dialog and any open modal behind it
- I don't love the black empty spots, I bet you could count the number of images and figure out a layout that fills all available space. 2, 3, 4, 6, 8, 9 are simple, 5 and 7 are a problem but you could probably do a 3 row and 2 rows.

staab:
> Have modified the note parsing, to aggregate images in a sub-array, this could also be handled in Welshman if it was to be normalised.

It would probably be good to add a `LinkGrid` to welshman as another ParsedType. Keep the logic out of `parse`, instead I think it would be better to add a new function similar to `truncate` called `reduceLinks` or something like that. This function could include the grid building logic too, so you'd get a data structure like this:

```
export type ParsedLinkGridValue = {
  links: ParsedLinkValue[][]
}
```

src/app/MenuDesktop.svelte
ticruz38: It used to be z-nav, I changed it to z-modal to have it above the modal backlight when rewriting the modals.
Changed it back to z-nav again while moving the modal backlight rightward because my new fullscreen image viewer would not get above this sidebar when triggered inside the modal.
Long story short it's a battle of z-index context. At the same time having the nav set to z-nav made sense

staab:
Nice, I like the solution you came up with. Another thing I noticed, can you fadethe images overlay in a bit?

ticruz38:
> Escape closes both the image dialog and any open modal behind it

Fixed by attaching keyboard event to the fullscreen container
> I don't love the black empty spots

The first image will take up the empty spots width if any
> It would probably be good to add a LinkGrid to welshman as another ParsedType. Keep the logic out of parse, instead I think it would be better to add a new function similar to truncate called reduceLinks or something like that

Added another type as well as a reduceLinks function in Welshman, this PR needs @welshman/content@0.016 to be published before testing

src/app/MenuDesktop.svelte
staab: The right thing to do here is probably to add a new z index or two in tailwind.config.js and use that, since the current state isn't correct.

staab (commented):
I'm still seeing escape close the background modal, and no transition on the image overlay:

https://github.com/user-attachments/assets/08bde6a4-9969-4d30-a2d7-ae21c3698989

ticruz38:
> The right thing to do here is probably to add a new z index or two in tailwind.config.js and use that, since the current state isn't correct.

adding new z-index won't help, if the fullscreen element triggers inside a fixed modal, it will inherit that z-index.
The dilemma is to have the modal go behind the menu, while the fullscreen should go above, which is not possible once the z-index is inherited from the modal.

ticruz38:
> I'm still seeing escape close the background modal

Weird, I'm default preventing the keyboard event bound to the fullscreen element. preventdefault was enough in my morning test to avoid the event bubbling to the modal, added stopPropagation now, can you try again?


https://github.com/user-attachments/assets/d874aece-99b7-43f2-b8c6-16b6279dca79

staab:
Escape and the transition are working fine now (although the transition should be only about 200ms, it's way too long right now)

> once the z-index is inherited from the modal.

The overlay is pretty simple, what if we used an entirely new component? It would still have to be integrated into history to support the back button. Probably preferable to fit it into the modal component (maybe a prop to control z index would work).

ticruz38:
I forced the fade transition duration to 200ms, 400 is the default.

> The overlay is pretty simple, what if we used an entirely new component? It would still have to be integrated into history to support the back button. Probably preferable to fit it into the modal component (maybe a prop to control z index would work).

I tried decoupling the overlay from the modal, it's enough to add one more z to tailwind and have the overlay and modal in a different html context. The problem arises when modals stacks up, the overlay is pushed behind all the modals, it should be in between them.

What does the z-nav break again? Is this the orange overlapping line? 

Screenshot 2025-01-29 at 11 59 47

staab:
> What does the z-nav break again? Is this the orange overlapping line?

Yeah, exactly

ticruz38:
I can still see it

staab:
- [ ] There should be some space between the grid and neighboring content (I believe we have this for links currently):
Screenshot 2025-01-29 at 12 32 22 PM

- [ ] Modals don't scroll while the overlay is open, but the regular page does
- [ ] It would be nice to avoid re-loading the image when opening the overlay:

https://github.com/user-attachments/assets/130fa4fb-4d4f-494c-b563-c002e2f35c71

- [ ] I can still see grids when I have media turned off
- [ ] X button and image click doesn't work if the image hasn't loaded yet
- [ ] Black placeholders look broken, show something less dark with a loading animation

ticruz38:
> There should be some space between the grid and neighboring content (I believe we have this for links currently)

Right, there is a my-2 css rules that wrap all links. Same for ImageGrid now, the ImageGrid respect the author layout, if there was one or more newline before or after the list of pictures, it is preserved.

> Modals don't scroll while the overlay is open, but the regular page does

Have added a noScroll property to the Carrousel that will prevent any scrolling behaviour.

> It would be nice to avoid re-loading the image when opening the overlay

Image used to watch the load event on the image via onMount, I'm using svelte built-in binding now and it seems to have fixed it. I have not seen any subsequent loading state when opening the image in zoom since.

> Black placeholders look broken, show something less dark with a loading animation

It is now a shimmer animation, as is the case on most website for image loading. It is now a greyish transparent color, it can be modified of course.

staab (commented):
We're so back:
Screenshot 2025-01-30 at 11 54 14 AM

On narrow screens, the shimmer placeholder is very tall:


Uploading Screen Recording 2025-01-30 at 12.00.30 PM.mov…

- [ ] Escape button is only sometimes working. Try navigating to the last image by clicking on the arrows, then hit escape.
- [ ] Could we also use the overlay instead of the modal for link previews and single links? We could simplify the NoteContentKind1 component and remove some dead code too if we did that.
- [ ] Add rounded corners to the images and shimmer placeholders in the overlay

ticruz38:
> Could we also use the overlay instead of the modal for link previews and single links? We could simplify the NoteContentKind1 component and remove some dead code too if we did that.

@staab What is the overlay element you are referring to here?

staab:
The dark background that opens when link grid is used. Right now every other image gets loaded in the drawer-style modal, which has never looked very good. We should be able to combine everything into fewer, simpler components. But don't worry about it right now, I took a look and it's annoyingly complex.

staab:
Merged via f68b5a9b

PR #550, opened 2025-01-23
Profile placeholder

ticruz38:
The profile picture placeholder now uses an fa-icon instead of the Coracle one.

Shows a loading gradient for 3 seconds and then stay still.

https://github.com/user-attachments/assets/4b75f94e-dc42-4f9d-9b02-c343090572df

### DIFF
--- a/src/app/shared/PlaceholderCircle.svelte
+++ b/src/app/shared/PlaceholderCircle.svelte
@@ -1,23 +1,27 @@
 <script lang="ts">
+  import {now} from "@welshman/lib"
   import cx from "classnames"
-  import {stringToHue, hsl} from "src/util/misc"
-  import LogoSvg from "src/partials/LogoSvg.svelte"
+  import {stringToHue, hsl, timestamp1} from "src/util/misc"
 
   export let pubkey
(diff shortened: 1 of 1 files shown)


staab:
😂
Screenshot 2025-01-23 at 10 18 19 AM

ticruz38:
Hmmm this is where vector graphic icon shines I guess

ticruz38:
https://github.com/user-attachments/assets/4fa9c8d6-8190-48af-b609-92d335807877

Screenshot 2025-01-24 at 13 45 56


- Using an SVG icons now as it automatically scale to its container. 
- The animation is also handled via SVG now. 
- The icon is from Nilliesmons npub149p5act9a5qm9p47elp8w8h3wpwn2d7s2xecw2ygnrxqp4wgsklq9g722q
- The SVG is not added to the Icon component, the requirement for this one differed too much from the others, especially the gradient and animation.

The overall look is MSNesque

staab:
Merged via ec896b77. I'm not sure I like the MSN association, but it's still an improvement over what we had before.

PR #551, opened 2025-01-23
Get rid of Hurdak

ticruz38:
Work in progress

Some hurdak (minor) functions do not yet have a equivalent in welshman, here is the list:

- ToTitle
- pluralize
- commaFormat (number to text formatting)
- quantify (commaFormat + pluralize)
- bytes (text to int formatting)
- round (Math.round, but support floating numbers too)
- filterVals (remove undefined keys in a map)

Let me know if I should add them to welshman or to a util folder in Coracle


staab:
- `bytes` can be replaced by the literal multiplication, since it only exists in one place.-
- `round` and `filterVals` can go into welshman
- The rest can be copied into util/misc

ticruz38:
there is another PR in welshman, @welshman/lib@0.0.38 needs to be published before merging

src/app/views/PublishesEvents.svelte
```diff
@@ -1,15 +1,15 @@
...
   import {LOCAL_RELAY_URL} from "@welshman/util"
   import {derived} from "svelte/store"
-  import {pluralize, seconds} from "hurdak"
   import Tile from "src/partials/Tile.svelte"
   import PublishCard from "src/app/shared/PublishCard.svelte"
+  import {pluralize} from "src/util/misc"
 
   $: recent = (Object.values($thunks) as Thunk[]).filter(
-    t => t.event.created_at > now() - seconds(24, "hour"),
+    t => t.event.created_at > now() - 24 * HOUR,
```
staab, line 12: This can just be `DAY`

src/app/views/PublishesEvents.svelte
```diff
@@ -42,7 +42,7 @@
   // If the page gets refreshed before pending finishes, it hangs. Set stuff to failed
   $: {
     for (const t of recent) {
-      if (t.event.created_at < now() - seconds(1, "minute")) {
+      if (t.event.created_at < now() - 1 * MINUTE) {
```
staab, line 45: no need to multiply by 1

src/util/misc.ts
```diff
@@ -54,17 +61,17 @@ export const formatTimestampAsDate = (ts: number) => {
...
-  if (delta < seconds(1, "minute")) {
+  if (delta < MINUTE) {
     unit = "second"
-  } else if (delta < seconds(1, "hour")) {
+  } else if (delta < HOUR) {
     unit = "minute"
-    delta = Math.round(delta / seconds(1, "minute"))
-  } else if (delta < seconds(2, "day")) {
+    delta = Math.round(delta / MINUTE)
+  } else if (delta < 2 * DAY) {
```
staab, line 69: This is fine, but for future reference I prefer to use `int(2, DAY)` which just makes things a little clearer

Issue #498, opened 2024-12-02
Improve mutes


staab:
- [ ] Allow users to hide the "show x hidden replies" bit for muted notes
- [ ] Don't mute notes by the user's own pubkey
- [ ] Hide muted parents in feeds (sometimes a reply to a muted note will show up, just remove it and its parent if the parent is muted).

nostr:nevent1qvzqqqqqqypzqnyqqft6tz9g9pyaqjvp0s4a4tvcfvj6gkke7mddvmj86w68uwe0qyghwumn8ghj7mn0wd68ytnvv9hxgtcpzamhxue69uhhyetvv9ujuurjd9kkzmpwdejhgtcppemhxue69uhkummn9ekx7mp0qyghwumn8ghj7mn0wd68ytnhd9hx2tcpzamhxue69uhk6mr9dd6jumn0wd68yvfwvdhk6tcqyr8kzdu4c9avfjyuxk8rmgzze8zf8tsed6watjjskf00sgc7rde5zjyzlap

ticruz38:
- The show x hidden replies only display when zooming on the note thread (in the modal). When showing replies, all of them are hidden by default. Do we remove that option in the modal as well?
- Note can be muted by user or by the user pubkey via the PersonActions component, should I remove this possibility?

staab:
> Don't mute notes by the user's own pubkey

By this I mean don't mute the user's own notes if they match a muted word. Keep the mute action and setting.

> The show x hidden replies only display when zooming on the note thread (in the modal). When showing replies, all of them are hidden by default. Do we remove that option in the modal as well?

Add a "hide_muted" setting to the content settings which doesn't factor in muted or low-wot notes into the "show x replies" button. The button will sometime still appear in the context of feeds, if not note detail pages.

ticruz38:
We need a screen to manage the muted profiles. I can't unmute the profile I muted for this PR

staab:
See Settings > Content

PR #553, opened 2025-01-28
Improve mute

ticruz38:
#498 

> Allow users to hide the "show x hidden replies" bit for muted notes

There is a new settings called "load_muted_replies", on by default, if turned off, it will hide the button

> Don't mute notes by the user's own pubkey

The only way to mute a note of your own is either to mute your own npub or that specific event id. If a word is muted and your note contain this word, it won't be hidden.

> Hide muted parents in feeds (sometimes a reply to a muted note will show up, just remove it and its parent if the parent is muted)

This one is the most challenging. It should be done in the new NoteReducer. 
The NoteReducer currently look for a parent at a depth of 2. If it turns out that the parent at the depth 3 is muted, we won't catch it.
The solution is obviously to loop until no parent is found just to check if one event in the chain is being muted, in which case all downstream event should be hidden. It might have a small performance footprint.

Awaiting your feedback on that last point to implement the needed changes

### DIFF
--- a/src/app/shared/FeedItem.svelte
+++ b/src/app/shared/FeedItem.svelte
@@ -9,7 +9,14 @@
   import {derived} from "svelte/store"
   import NoteMeta from "src/app/shared/NoteMeta.svelte"
   import Note from "src/app/shared/Note.svelte"
-  import {ensureUnwrapped, getSetting, isEventMuted, loadEvent, sortEventsDesc} from "src/engine"
+  import {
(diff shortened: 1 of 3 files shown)


staab:
>  If it turns out that the parent at the depth 3 is muted, we won't catch it.

That's ok, we don't have to get too crazy, we should just avoid showing immediate descendants of muted notes if we know we've muted the parent. Deep descendants should still be fair game.

Issue #547, opened 2025-01-22
Show zaps with a message


staab:
The tricky thing is that lots of messages are boilerplate. Maybe we can do this, maybe not.

ticruz38:
Show the message only for the largest zap? All zaps? Under the note? In a feed?

staab:
Run this in the console to get a list of zap messages by how many times they show up:

```
groupBy(identity, repository.query([{kinds: [ZAP_RESPONSE]}]).map(e => e.content || JSON.parse(e.tags.find(t => t[0] === 'description')[1]).content).filter(identity))
```

I was originally thinking of showing these as replies. But most appear to be either emojis or ads. Since we don't show reaction emojis anyway, there's no need to show zap emojis. Other messages don't seem to add much, but it's a shame to allow users to provide messages, but then ignore them. So I think the best way to solve this would be to address this on the notifications page via NoteMeta (see recent update).

This means that this should be pretty simple; the zap section in NoteMeta will just have to be updated to show zap content. While you're at it, we might as well show a full summary there rather than just the first zap/repost/reaction. Here's what it currently looks like:

<img width="659" alt="Image" src="https://github.com/user-attachments/assets/84daadd6-9fa9-4652-a9bf-cb0d99de3075" />

This is sort of weird, because it doesn't say what the other 15 people did. Of course, showing all activity might make NoteMeta too unwieldy in a feed context, so it's possible we'll want to add a `mode` flag to NoteMeta, or create a separate component.

PR #557, opened 2025-01-29
Show richer zap reactions #547

ticruz38:
In the notification tab:

If there is more than one zap, show all zap activity and aggregate similar messages (many are boilerplate) into chips containing the message and aggregated people's avatar. 
Same thing for reactions.
Screenshot 2025-01-29 at 16 48 16

If it is a one person action, show the zap amount and content as well, but also the date of the event

Screenshot 2025-01-29 at 17 05 09

Most of the logic happens in the NoteActivity component, it handles reactions and interactions and differ quite a bit from NoteMeta.


staab:
- [ ] Try to reduce the redundant layout code (date should always be in the same place, at the same font size
- [ ] Zaps aren't correct, you're using the same zapper for all zaps, even from different people
- [ ] Given that interactions/replies kinds share very little code, go ahead and separate the two
- [ ] Remove borders and make the border radius for the pills `rounded-full`
- [ ] Use the same pill style for zaps by multiple people as for reactions
- [ ] Make pills clickable, opening a modal with the list of people in it (see PeopleAction for some code to copy)

ticruz38:
> Zaps aren't correct, you're using the same zapper for all zaps, even from different people

It's a copy/paste from zaps in NoteActions. My understanding is that all events (here context) zapping the same root event have the same zapper (the root event zapper). It seems to be working well, but my test surface for zaps is low.

> Try to reduce the redundant layout code (date should always be in the same place, at the same font size

I could not reuse exactly the same layout as in NotificationItem for NoteReactions. It is a multiline component, so one different date and different formatting for zap and reaction. The code could be reused but that would involved a bunch of slot and/or props, it looks similar but there is a bunch of difference even in the layout.

src/app/shared/NotificationItem.svelte
```diff
@@ -25,20 +25,11 @@
...
-        actionText="{verb} to a note mentioning you" />
-    {/if}
-    <small>{formatTimestamp(max(pluck("created_at", [event, ...getContext(event)])))}</small>
-  </div>
+<NoteReducer {events} depth={1} bind:items let:event let:getContext>
+  {#if kind === "reactions"}
+    <NoteReactions context={getContext(event)} {event} />
+  {:else}
+    <NoteInteractions context={getContext(event)} {event} />
+  {/if}
```
staab, line 33: The kind parameter has become a little goofy, let's just make this a slot instead and pass in NoteReducer and the correct child component directly from the caller.

staab (commented):
- [x] Give the pills a little margin (probably px-2 py-1, or maybe just p-1)
- [x] Only show a single date for each group of reactions/zaps/reposts
- [ ] Use text-sm to make the description text the same size as the date

ticruz38:
Implemented the suggested changes, the date looks out of place in certain conditions, here when multiple zaps are in, it shrink the zap container, while anything below (here the likes) will take the full width. There is no easy "fix" for this

Screenshot 2025-01-31 at 13 12 08

src/app/shared/NotificationItem.svelte
```diff
@@ -25,20 +22,9 @@
...
-    {:else}
-      <PeopleAction
-        pubkeys={uniq(pluck("pubkey", getContext(event)))}
-        actionText="{verb} to a note mentioning you" />
-    {/if}
-    <small>{formatTimestamp(max(pluck("created_at", [event, ...getContext(event)])))}</small>
-  </div>
+<NoteReducer shouldAwait shouldSort {events} depth={1} bind:items let:event let:getContext>
+  <slot {event} context={getContext(event)}>
+    <div>Missing child component</div>
```
staab, line 27: No need for an error, just use a self-closing slot tag

staab (commented):
I agree, let's just remove the date entirely since we already have the day showing above.

staab:
Merged via 524922fb

Issue #548, opened 2025-01-22
Add abortController to feed loaders


staab:
So that when we change the feed or navigate away we can stop loading stuff.

ticruz38:
The signal I am passing via the request is being lost once the subscription inevitably goes through optimizeSubscriptions.

For feeds, my guess is that optimizeSubscriptions is not necessary so we should probably use an additional parameter in the request to signify that this particular sub must not be concatenated with others and keep its own abortcontroller.

ticruz38:
In the stack trace, it's instantiating 2 same Subscriptions, the first one go through optimizeSubscription and is dropped for another but the request object is modified with new filters and no signal

ticruz38:
> Is that not happening? Or maybe we need to propagate that to the controller's signal

Absolutely, this is happening. I got distracted by the fact that my subscription was never executed. I missed this part where my signal aborts another signal.

staab:
You can skip this one, I'm still thinking about how to approach fixing subscribe

PR #560, opened 2025-02-03
Abort signal for feeds #548

ticruz38:
Use an abort signal in the feed request that will abort on each feed selection and upon page leave.

- The random key is replaced by the feed identifier.
- The key check in the onEvent callback of the feed controller is removed, as abort should prevent any more events from being streamed.


staab:
Merged via 54ffdae3

PR #567, opened 2025-02-09
Started svelte-i18n integration

papiche:
Add language selector using [svelte-i18n](https://github.com/kaisermann/svelte-i18n)
This MR is to inform about "work in progress"

When used in code, it replaces fixed text with dynamic

For example, in ```./src/app/MenuDesktop.svelte```
```Settings``` becomes ```{$_('page.home.settings.title', { default: 'Settings' })}```

according to  ```./src/app/localization/lang/LANG.json```

Before translating, this needs to get widely applied to populate "en.json" with all sentences.
Is it OK for you ? How would you like we get organized ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,5 @@ public/images/maskable-icon-512x512.png
 public/images/pwa-*
 android/app/release/*
 icons/*
+.aider*
+.env
--- a/index.html
+++ b/index.html
@@ -42,7 +42,7 @@
   </head>
   <body class="w-full">
     <div id="app"></div>
-    <script type="module" src="/src/main.js"></script>
+    <script type="module" src="./src/main.js"></script>
     <script defer data-domain="coracle.social" src="https://plausible.coracle.social/js/script.manual.js"></script>
   </body>
 </html>
--- a/package-lock.json
+++ b/package-lock.json
@@ -2414,12 +2414,23 @@
       }
     },
     "node_modules/@eslint/plugin-kit": {
-      "version": "0.2.2",
-      "resolved": "https://registry.npmjs.org/@eslint/plugin-kit/-/plugin-kit-0.2.2.tgz",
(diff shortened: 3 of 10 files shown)

### CODE REVIEW

**Comment by staab** [comment]
This looks quite good, thank you! I will likely move the language selector somewhere less visible ultimately. My one question is whether it would make sense for commonly used terms (like "Relay") to get their own key? Probably premature abstraction, but I don't know what best practices are here.

**Comment by papiche** [comment]
> would make sense for commonly used terms (like "Relay") to get their own key

As mean can change from context and synonyms can be used, it is better to make translation for each application routes like :

```
{
"page": {
  "home": {
    "feeds": "Feeds",
    "relays": "Relay",
    ... },
  "notes": { ...
    },
  "settings": { ...
}
```

I must still explore and discover Coracle to be able to make a complete i18n json
To get help, i'm trying to get in touch with @Cristol-7 (https://github.com/coracle-social/coracle/issues/561) ...

Let's see how it goes ;)

File: .env [hunk 1]
```diff
@@ -1,27 +1,27 @@
...
+VITE_APP_URL=https://coracle.copylaradio.com
 VITE_APP_LOGO=/images/logo.png
-VITE_APP_WORDMARK_DARK=/images/wordmark-dark.png
-VITE_APP_WORDMARK_LIGHT=/images/wordmark-light.png
-VITE_APP_DESCRIPTION="A nostr client for regular people"
-VITE_CLIENT_NAME=Coracle
+VITE_APP_WORDMARK_DARK=/images/miz.png
+VITE_APP_WORDMARK_LIGHT=/images/miz.png
+VITE_APP_DESCRIPTION="A nostr client to emancipate people freedom"
+VITE_CLIENT_NAME=CoracleZ
```
staab, line 24: Instead of modifying .env, copy it to .env.local and modify it there

PR #580, opened 2025-02-17
I18n tools

Sananda-KUMARA:
Proposed translation tool for Coracle UI internationalization (i18n): Create, Delete, Update (CRUD) of language keys and text fields for filling translation string literals. Option to export in indented JSON format (nested keys). Note: keys are stored in a Svelte i18n store after loading the 4 currently supported language files.

capture-1
capture-2


staab:
I'd prefer not to build the translation editor into the tool. Aren't there existing tools for managing translations? Also, the diff is a mess, it looks like your editor changed formatting a bunch.

PR #626, opened 2025-06-25
add tooltip in qrcode of login bunker

lucaskbr:
I was learning about the app and nostr

And when i tried to connect the coracle with nsec i was a little bit lost
coracle was showing a qr code and nsec was showing a input field, at the time i didnt saw the cursor changing to copy when i hovered the qr code

I think adding a tooltip in the qr code will help in the UX

### DIFF
--- a/src/app/views/LoginBunker.svelte
+++ b/src/app/views/LoginBunker.svelte
@@ -10,6 +10,7 @@
   import Anchor from "src/partials/Anchor.svelte"
   import FlexColumn from "src/partials/FlexColumn.svelte"
   import Heading from "src/partials/Heading.svelte"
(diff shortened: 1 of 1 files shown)


lucaskbr:
I wasnt able to run master BTW

staab:
Looks good, thanks! What problems did you run into running it?

staab:
Oh, if it was dev then that makes sense, it's on an in-development version of welshman right now.

staab:
Got it, bunker connections are kind of arcane still. The qr code is a nostrconnect:// string, and the bunker url nsec.app provides can be pasted into the input. But everyone could do a better job explaining that.

Issue #586, opened 2025-03-01
UI issue when replying to notes


Dikaios1517:
When replying to notes and pasting long, unbroken, text strings, such as for content addressable images, the "post" icon gets shifted to the center of the composition window. Additionally, the user is no longer able to interact with it, either by clicking the location it shifted to, or its original position.

![Example Image](https://relay.brightbolt.net/f5a3858b85e8b31a80ad431c7c521414f0b200d605d8b8c94e59a77920a1f2e4.jpg)

LeParrain0:
Fix: src/app.css, .tiptap block (line 352, before the closing }):

===================
  overflow-wrap: anywhere;
===================

Why: the editor and send button are flex siblings. A long space-less string won't wrap (pre-wrap only breaks on spaces), so it blows up the editor's min-width, pushes the button off, and overflow-x-hidden hides it. anywhere (not break-word) also shrinks min-content width, so the layout can collapse again.

Can i PR ?

staab:
Please do, I've tried to fix this a few times. I'm not sure word-wrap: anywhere is great though, since it will split words that shouldn't be split.

LeParrain0:
Done. You can check

PR #666, opened 2026-07-27
Fix send button shifting off-screen on long unbroken text (#586)

LeParrain0:
When replying and pasting a long, unbroken string (e.g. a content-addressable image URL), the send button was pushed toward the center of the composer and became unclickable.

Cause: the editor (TipTap/ProseMirror) and the send button are flex siblings. white-space: pre-wrap only breaks on spaces, so a space-less string is one unbreakable "word" with a huge min-content width. That stops min-w-0 from shrinking the editor, so it overflows and overflow-x-hidden clips the button out of reach.

Fix: add overflow-wrap: anywhere to .tiptap in src/app.css. Unlike break-word, anywhere also reduces the element's min-content width, letting the flex/grid layout collapse. Same approach already used in MediaLinkPreview.svelte.

### DIFF
--- a/src/app.css
+++ b/src/app.css
@@ -349,6 +349,7 @@ body,
   --tiptap-object-fg: var(--neutral-800);
   --tiptap-active-bg: var(--neutral-200);
   --tiptap-active-fg: var(--neutral-800);
+    overflow-wrap: anywhere;
 }
 
 .tiptap-suggestions {


staab:
The selector wasn't quite specific enough, but the fix works. I've released it as 6f84b3ebafa99d48379d57cb7a73d50a164088e3, thanks!

LeParrain0:
You're welcome. Is there any way to continue to contribute to that awesome project and get paid ?

staab:
No, sorry, I don't have a budget for contributors right now.

LeParrain0:
Totally understandable, and no pressure at all. Just so you know, my rates are very modest and completely flexible — so if a small contribution ever fits your budget, even a few dollars, I'd be more than happy to keep helping out. And if not, I'll gladly keep contributing anyway because I really enjoy the project. Either way, thanks for building something this cool!

PR #668, opened 2026-08-27
Fix: Invert close and back button behavior

### LINKED ISSUE #576: minor UI nitpick
the functionality of the bright orange X button in the top right corner (essentially functions as a back button) and the understated grey down arrow below it (closes all tabs and returns to feed) feel like they should be reversed

jrsd86:
## Summary
Fixes the incorrect button behavior in modal navigation.

- The X button (orange) now closes all modals as expected
- The arrow button (gray) now goes back one page as expected

## Related Issue
Fixes #576


staab:
It already works as intended.

PR #669, opened 2026-08-27
feat: Add toggle to show/hide reactions

### LINKED ISSUE #554: Show reactions
Make this a user setting so people can opt in. Possibly show zaps as well if this is enabled #547

jrsd86:
## Summary
Adds a user setting to toggle the display of reactions on notes.

- Users can disable reaction display in settings
- Reactions are hidden in both NoteReactions and NoteMeta components
- Defaults to enabled for existing users

## Related Issue
Fixes #554


staab:
This setting already exists as note_actions. The change to NoteMeta also is incorrect, it should be solved by whatever is feeding the events to that component.

PR #672, opened 2026-09-12
feat: add a separate setting for link previews

### LINKED ISSUE #622: Ability to show the images but not link previews
There is a setting called `Show images and link previews`, but I only want it to display images, not link previews. Instead, I’d like links to appear as plain hyperlinks while images still render as images.

jrsd86:
## Summary

`show_media` currently controls both embedded media and link previews, so
there is no way to keep media visible while hiding preview cards. This adds
a separate `show_link_previews` setting. When disabled, links render as plain
hyperlinks while media continues to render normally. Both settings default to
enabled, preserving the current behavior.

## Changes

- Add `show_link_previews` to `defaultSettings`, defaulting to `true`
- Add a "Show link previews" toggle to content settings, and relabel the
  existing toggle to "Show media" since it no longer covers previews
- Pass the setting from `MediaGrid` down to `Media`, which renders a plain
  hyperlink instead of a preview card when it is disabled

## Testing

- `pnpm run check` and `pnpm run test:unit` pass
- Rendered `Media` against image, video, audio and plain link URLs with the
  setting both on and off. Only the plain link case changes; images and
  embedded media are unaffected in both states.

## Related Issue

Fixes #622


staab:
Thanks, looks good!

